PR #1, opened 2019-08-21
Updates from switch to Store Codec Pack version

mattrwoz:
This document has not been updated since we switched HEVC to use a store codec pack.
* CoCreateInstance no longer works when using store codec pack.
* Support for DXVA2 has been removed
* mshevcdec.dll has never been the dll name, it has always been hevcdecoder.dll
* hevcdecoder_store.dll is used in the codec pack


PRMerger9:
@mattrwoz : Thanks for your contribution! The author(s) have been notified to review your proposed change.

drewbatgit (approved):

drewbatgit:
#sign-off

PR #7, opened 2019-08-25
fix visibility of <NUL>

yoni1:
The <NUL> is not html-escaped, so it is invisible in the rendered docs.


PRMerger6:
@yoni1 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

knicholasa:
Thanks for catching this and making a contribution!

PR #14, opened 2019-08-30
Fix table formatting which had merged string & macro

rbroker:
The table formatting in `enabling-and-disabling-privileges-in-c--.md` had merged the macro name and string representation. It looks like the table really should have 3 columns, to bring it a bit more in line with the definitions from `winnt.h`

Thanks,
Richard


PRMerger6:
@rbroker : Thanks for your contribution! The author(s) have been notified to review your proposed change.

knicholasa:
Looks good to me. Thanks for catching this!

PR #15, opened 2019-08-31
Fix incorrect information regarding NonUniformResourceIndex

AdamJMiles:
This information was incorrect. It is only within a wave that the index is not allowed to vary, not within a draw/dispatch.

### DIFF
--- a/desktop-src/direct3d12/resource-binding-in-hlsl.md
+++ b/desktop-src/direct3d12/resource-binding-in-hlsl.md
@@ -83,7 +83,7 @@ sampler samp[7] : register(s0);
 tex1[myMaterialID].Sample(samp[samplerID], texCoords);
 ```
 
-There is an important default restriction on the use of the indices (`myMaterialID` and `samplerID` in the code above) in that they are not allowed to vary within a wave (aka 'wavefront' or 'warp'). Even changing the index based on instancing counts as varying.
+There's an important default restriction on the use of the indices (`myMaterialID` and `samplerID` in the code above) in that they're not allowed to vary within a [wave](/windows/win32/direct3dhlsl/hlsl-shader-model-6-0-features-for-direct3d-12#terminology). Even changing the index based on instancing counts as varying.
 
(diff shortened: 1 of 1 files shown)


PRMerger20:
@AdamJMiles : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims:
Hi, Adam, what's the provenance of the change you're proposing? Does it come from the product team? And, can a developer actually take advantage of the distinction in granularity that you're calling out? Specifically, can the developer write code in such as way as to guarantee that a change to one or more of those indices will fall between two distinct waves, and if so how? If they can't (if the next draw/dispatch call actually turns out to be their earliest safe opportunity for changing those indices), then it might be better to leave the info as-is and not introduce unnecessary details. Thanks! ~Steve

stevewhims (changes requested):
Hi, Adam, what's the provenance of the change you're proposing? Does it come from the product team? And, can a developer actually take advantage of the distinction in granularity that you're calling out? Specifically, can the developer write code in such as way as to guarantee that a change to one or more of those indices will fall between two distinct waves, and if so how? If they can't (if the next draw/dispatch call actually turns out to be their earliest safe opportunity for changing those indices), then it might be better to leave the info as-is and not introduce unnecessary details. Thanks! ~Steve

ajmiles:
Hi Steve,

The provenance of the change is my own understanding of the D3D12 API and Shader Model 6.

Developers can take advantage of the finer granularity in Compute Shaders. If a particular thread group is sized to match the GPU's natural wave size (which can be queried from both within D3D12 and inside HLSL itself) then it's possible to take a different code path where uniformity is known to be guaranteed across the wave. With the addition of Shader Model 6 Wave Intrinsics the concept of a 'wave' has been codified in the design such that developers can do wave-level operations such as WaveBallot, WaveMin, WaveMax etc, so the concept is now well defined and is something developers can take advantage of in their own shaders.

PRMerger17:

stevewhims (approved):
Sounds good! Approved!

PR #53, opened 2019-09-20
Dot-local redirection requires reg key edit

amyw-msft:
Dot-local redirection requires that DevOverrideEnable is set to 1 in order to work, and this page fails to mention that. That said, I'm not sure if this is what we should recommend since it does require global registry edits? App-local redirection is usually the best way to do this - perhaps there should be two pages or two sections? I'm not familiar with the "normal scenario" for dot-local redirection - I mostly use it to test OS DLLs that are linked to with APISETs for which app-local redirection does not work, but that's not a normal customer scenario.


PRMerger9:
@stwish-msft : Thanks for your contribution! The author(s) have been notified to review your proposed change.

mcleanbyron (approved):

PR #60, opened 2019-09-27
Fix use of special HTML characters

henrygab:
Fix less-than (`<`) and greater-than (`>`) symbols to be properly HTML-escaped.

Without this fix, the HTML version of the pages as viewed on docs.microsoft.com fail to properly show all the operators.


PRMerger12:
@henrygab : Thanks for your contribution! The author(s) have been notified to review your proposed change.

mcleanbyron (approved):

mcleanbyron:
#sign-off

PR #95, opened 2019-10-18
Added hexcode to the Standard access rights

trir262:
I was looking for a summary of the hex codes of every access right, including the standard ones. Since I did not find it, I experimented myself and found these codes to be correct on a windows 2016 server.


PRMerger9:
@trir262 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

mcleanbyron (approved):

PR #109, opened 2019-10-30
Correct export directory example code

MauriceKayser:
The example code is wrong and does not work, see [this](https://stackoverflow.com/a/20453007) answer from 2013 and [this](https://stackoverflow.com/a/40001778) from 2016. Tried to make the explanation more clear and added an example how to search for an RVA by ordinal instead of by name.
The corrected algorithm yields the same results as `GetProcAddress(name)` and `GetProcAddress(biased_ordinal)` do.
Not sure if `ms.date` meta data has to be changed by pull requests.

### DIFF
--- a/desktop-src/Debug/pe-format.md
+++ b/desktop-src/Debug/pe-format.md
@@ -1673,26 +1673,35 @@ An export name is defined only if the export name pointer table contains a point
 
 #### Export Ordinal Table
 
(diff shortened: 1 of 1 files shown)


PRMerger20:
@mauricekayser : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Karl-Bridge-Microsoft (approved):
@mauricekayser - thanks for correcting this!

Karl-Bridge-Microsoft:
#sign-off

PR #112, opened 2019-10-31
Fix x64 crash due to pointer truncation

edwinvp:
The code stores a pointer in the window that points to the DemoApp object instance (using SetWindowLongPtrW). The idea is to retrieve the object pointer later when it is needed for processing window message. Conversion to unsigned long is done by PtrToUlong and this unfortunately truncates the 64-bit pointer to a 32-bit unsigned long before SetWindowLongPtrW sees it. The corrupted pointer is then later used and at that point the program crashes on most but not all machines (ASLR?).


PRMerger17:
@edwinvp : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims:
Thanks, @edwinvp! It's now safe for you to delete your branch.

PR #130, opened 2019-11-14
Add "streamid" field

AlexLaforge:
The "streamid" field is present in Windows Server 29016, but documented ANYWHERE. This should change, as a lot of developers require this information.


PRMerger6:
@AlexLaforge : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (approved):
Thanks, I'll approve! I don't like the question marks, but this is at least an improvement, so maybe we can loop back later to complete it.

stevewhims:
It's now safe for you to delete your working branch.

PR #137, opened 2019-11-15
Update -wic-native-image-format-metadata-queries.md

wstaelens:
fixing slashes at the end because the proposed alternative query path doesn't end with a forward slash , so the original should also not end with one. either both end with a slash (original and alternative query path) or both don't use the forward slash.


PRMerger16:
@wstaelens : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (approved):
Agreed: thanks for the fix!

PR #142, opened 2019-11-18
Fix tag names as inline code in VML docs

hteumeuleu:
Throughout all the "How to use VML on Web Pages" docs, VML and HTML tag names used as inline code are being rendered as HTML in a real browser. This PR should fix that by marking every tag name as an inline code.


PRMerger16:
@hteumeuleu : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (approved):
Looks good, thanks for your hard work here!

stevewhims:
You can now safely delete your personal branch.

PR #144, opened 2019-11-20
Update schannel.md

xiaoyinl:
Delete an incorrectly encoded character after "public-key".


PRMerger7:
@xiaoyinl : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #145, opened 2019-11-26
Remove internal-only references in Windows version compatibility discussions; otherwise improve same

metathinker:
These documentation pages discuss how to mark your app as compatible with Windows 8.1 and Windows 10, as well as the "version lie by default" compatibility change in Windows 8.1 which requires many apps to explicitly mark themselves as compatible with newer versions of Windows if they don't want to be told that they are running on Windows 8.

The previous style and content of these pages was a little out of step with the rest of Windows docs, as they were originally based on an internal document that wasn't meant for customer-ready documentation. In particular, they make references to `sources` files and C preprocessor macros which only exist in Windows' internal build system ("Razzle").

The updates I've made are as follows:
- Delete all Razzle-ish references and usages or replace them with constructs everyone can use.
- Delete the `<trustInfo>` section from all but one of the example app manifests. Clarified why `<trustInfo>` is "recommended" - the previous text implied that it is important to OS versioning, but in fact it is useful to bypass a _different_ compatibility behavior (UAC file system virtualization).
- Fixed some formatting and updated some links which still work but redirect.

### DIFF
--- a/desktop-src/w8cookbook/operating-system-version-changes-in-windows-8-1.md
+++ b/desktop-src/w8cookbook/operating-system-version-changes-in-windows-8-1.md
@@ -14,6 +14,7 @@ ms.assetid: 3040262A-85EB-4F26-BE34-D2BBD5886E9E
 ## Platforms
 
 **Clients -** Windows 8.1
+
 **Servers -** Windows Server 2012 R2
 
 ## Description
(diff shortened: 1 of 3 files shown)


PRMerger19:

desktop-src/w8cookbook/operating-system-version-changes-in-windows-8-1.md
```diff
@@ -10,116 +10,93 @@ ms.date: 05/31/2018
```
metathinker, line 10: This documentation page is substantially duplicated in 2 places:
- https://docs.microsoft.com/en-us/windows/win32/w8cookbook/operating-system-version-changes-in-windows-8-1 (this file)
- https://docs.microsoft.com/en-us/windows/compatibility/operating-system-version-changes-in-windows-8-1

Unfortunately, the other copy of the source file isn't open-source on GitHub as far as I can tell, though I do have internal access to it. I'll submit an internal PR to update it, but the 2 files should really be merged or one deleted altogether.

desktop-src/w8cookbook/windows-version-check.md
```diff
@@ -23,75 +23,49 @@ Some apps perform a version check and simply pass a warning to users. However, t
 -   If the app is dependent on specific API functionality, ensure you target the correct API version.
```
metathinker, line 23: This documentation page is substantially duplicated in 2 places:
- https://docs.microsoft.com/en-us/windows/win32/w8cookbook/windows-version-check (this file)
- https://docs.microsoft.com/en-us/windows/compatibility/windows-version-check

Unfortunately, the other copy of the source file isn't open-source on GitHub as far as I can tell, though I do have internal access to it. I'll submit an internal PR to update it, but the 2 files should really be merged or one deleted altogether.

mcleanbyron (approved):

mcleanbyron:
#sign-off

PR #146, opened 2019-11-26
Update ldap-adspath.md

johanbenschop:
The docs only showed () and not (\\) because the backslash also needs to be escaped in markdown.


PRMerger6:
@johanbenschop : Thanks for your contribution! The author(s) have been notified to review your proposed change.

hkhmsft:
Hi @johanbenschop , thanks for your contribution, I've merged your changes and they should be live soon!

PR #164, opened 2019-12-23
Fix define syntax

gustavolev:
Fix define syntax in enabling-an-assembly-in-an-application-without-extensions.md


PRMerger12:
@iGustL : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #169, opened 2020-01-02
Update LVM_SETGROUPINFO docs

hughbe:
See https://github.com/dotnet/winforms/issues/2554 and http://users.skynet.be/oleole/Listview_grouping_feature.htm

- Fix dead link
- Update documentation to indicate that `LVGF_GROUPID` should not be set in order to update `LVGROUPW` properties

/cc @weltkante

### DIFF
--- a/desktop-src/Controls/lvm-setgroupinfo.md
+++ b/desktop-src/Controls/lvm-setgroupinfo.md
@@ -18,17 +18,17 @@ ms.date: 05/31/2018
 
 # LVM\_SETGROUPINFO message
 
-Sets group information.
+Sets group information. Send this message explicitly or by using the [**ListView\_SetGroupInfo**](/windows/desktop/api/Commctrl/nf-commctrl-listview_setgroupinfo) macro.
 
 ## Parameters
 
 <dl> <dt>
 
-*wParam* 
+*wParam* \[in\]
 </dt> <dd>ID that specifies the group whose information is to be set.</dd> <dt>
 
-*lParam* 
-</dt> <dd>Pointer to an <a href="/windows/desktop/api/Commctrl/ns-commctrl-taglvgroup">**LVGROUP**</a> structure that contains the information to set.</dd> </dl>
+*lParam* \[in, out\]
+</dt> <dd>Pointer to a [**LVGROUP**](windows/win32/api/commctrl/ns-commctrl-lvgroup) structure that contains the information to set.</dd> </dl>
 
 ## Return value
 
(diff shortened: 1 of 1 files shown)


PRMerger7:
@hughbe : Thanks for your contribution! The author(s) have been notified to review your proposed change.

desktop-src/Controls/lvm-setgroupinfo.md
```diff
@@ -28,19 +28,19 @@ Sets group information.
...
-</dt> <dd>Pointer to an <a href="/windows/desktop/api/Commctrl/ns-commctrl-taglvgroup">**LVGROUP**</a> structure that contains the information to set.</dd> </dl>
+</dt> <dd>Pointer to an <a href="https://docs.microsoft.com/en-us/windows/win32/api/commctrl/ns-commctrl-lvgroup">LVGROUP</a> structure that contains the information to set.</dd> </dl>
 
 ## Return value
 
 Returns the ID of the group if successful, or -1 otherwise.
 
 ## Remarks
 
+<b>LVGROUP.mask</b> should not include the <b>LVGF_GROUPID</b> flag when updating fields such as <b>pszTask</b> or <b>uAlign</b> on an existing group.
```
weltkante: Sounds too generalized to me, you can include LVGF_GROUPID as long as you actually change the ID. I think its better to document that this flag is a change-id request and will fail if the target group id already exists. How about something like
```suggestion
Including <b>LVGF_GROUPID</b> in <b>LVGROUP.mask</b> implies changing the group ID and will fail if the ID given in <b>LVGROUP.iGroupId</b> is already used. When updating fields such as <b>pszTask</b> or <b>uAlign</b> on an existing group including <b>LVGF_GROUPID</b> in <b>LVGROUP.mask</b> is not required.
```

desktop-src/Controls/lvm-setgroupinfo.md
```diff
@@ -28,19 +28,19 @@ Sets group information.
 </dt> <dd>ID that specifies the group whose information is to be set.</dd> <dt>
 
 *lParam* 
-</dt> <dd>Pointer to an <a href="/windows/desktop/api/Commctrl/ns-commctrl-taglvgroup">**LVGROUP**</a> structure that contains the information to set.</dd> </dl>
+</dt> <dd>Pointer to an <a href="https://docs.microsoft.com/en-us/windows/win32/api/commctrl/ns-commctrl-lvgroup">LVGROUP</a> structure that contains the information to set.</dd> </dl>
```
RussKie: We must not include locale information so users view docs in their language:
```suggestion
</dt> <dd>Pointer to an <a href="/windows/win32/api/commctrl/ns-commctrl-lvgroup">LVGROUP</a> structure that contains the information to set.</dd> </dl>
```

desktop-src/Controls/lvm-setgroupinfo.md
RussKie: How about something like this?
```suggestion
To change a group ID of an existing group add <b>LVGF_GROUPID</b> to <b>LVGROUP.mask</b> and set <b>LVGROUP.iGroupId</b> to the new ID. The call will fail if <b>LVGROUP.iGroupId</b> contains ID of an existing group.<br />
To update other properties of an existing group (e.g. update an alignment of the header or footer text for the group, <b>uAlign</b>) <b>LVGROUP.mask</b> must not contain <b>LVGF_GROUPID</b>, else the update will fail.
```

hughbe:
Thanks for the suggestions!

PRMerger19:
@hughbe : Thanks for your contribution! The author(s) have been notified to review your proposed change.

hkhmsft:
Thanks for your contribution, I've merged your changes and they should be live soon.

PR #177, opened 2020-01-12
Fix unescaped anchor tag running loose

atakiya:
Replace literal angled brackets of the **unescaped** anchor tag with appropiate character entities as else this mess happens on the MSFT Docs page
image


PRMerger6:
@Avunia : Thanks for your contribution! The author(s) have been notified to review your proposed change.

hkhmsft:
Thanks for your contribution, I've merged your changes and they should be live soon.

PR #207, opened 2020-02-05
RWByteAddressBuffer and Interlocked functions available on all stages

AdamJMiles:
A remnant of when D3D only allowed UAVs on PS and CS stages, that's no longer the case so these intrinsics work on all shader stages


PRMerger20:
@AdamJMiles : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (approved):
Looks good, thank you!

PR #214, opened 2020-02-11
Wrong example code

phuclv90:
- There's no `Shell_Windows` in [Shell](https://docs.microsoft.com/en-us/windows/win32/shell/shell)
- There's no `alert` in VBS

The above examples are also incorrectly copied to the [`_NewEnum` documentation](https://docs.microsoft.com/en-us/windows/win32/shell/shellwindows--newenum-method-7ral) even though there's no `_NewEnum` call in the code. I don't know how to raise an issue for that because there's no "Feedback" button so I'm putting it here

### DIFF
--- a/desktop-src/shell/shellwindows-item.md
+++ b/desktop-src/shell/shellwindows-item.md
@@ -63,7 +63,7 @@ JScript:
         var objShell = new ActiveXObject("shell.application");
         var objShellWindows;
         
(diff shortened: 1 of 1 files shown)


PRMerger13:
@phuclv90 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

mcleanbyron (approved):

PR #267, opened 2020-03-16
Update some non-working Markdown to correctly link.

gconnell:
There's a mix of markdown and HTML here, and some of the markdown within the HTML is not being HTML-ified.  This fixes a few of those links.  I haven't comprehensively checked for others, either in this file or elsewhere in the docs.


PRMerger17:
@gconnell : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #309, opened 2020-04-13
Update about-update-assessor-service.md

agoo113:
This API is designed for admin access only. This information needs to be added to the document so that the users are clear about the usage.


PRMerger6:
@agoo113 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

mcleanbyron (approved):

PR #312, opened 2020-04-15
Update principal-displayname.md

rbleattler:
Fixed Typo "For example, the setting this property value to..." to "For example, setting this property value to".


PRMerger20:
@rbleattler : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #313, opened 2020-04-16
Fix broken list due to mismatching dl tags, update MSDN links

phuclv90:
There are more closing `</dl>` tags than opening ones (72 vs 61). In fact if a property has a list of values then it'll miss a `<dl>` tag. That makes the value list failed to be rendered and indented properly. It's impossible to differentiate between properties and their values, both are bold in the same font face/size and printed in the same indentation

image

I've also fixed small errors like the broken ARM above, and updated the old msdn links to the redirected links too. The format may need some further refinement but now this list is completely readable compared to the previous version

### DIFF
--- a/desktop-src/CIMWin32Prov/win32-processor.md
+++ b/desktop-src/CIMWin32Prov/win32-processor.md
@@ -75,9 +75,9 @@ api_location:
 - CIMWin32.dll
 ---
 
-# Win32\_Processor class
+# Win32_Processor class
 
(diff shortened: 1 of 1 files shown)


PRMerger14:
@phuclv90 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (approved):
Thank you! :)

PRMerger6:

PR #316, opened 2020-04-16
Update using-color-in-direct2d.md

amrojjeh:
- Added missing symbols
- Put RGB and alpha values in bold so that they're easy to reference back to when stepping through the calculations
- Removed the dl element since that was invalidating the newlines


PRMerger16:
@dinglydo : Thanks for your contribution! The author(s) have been notified to review your proposed change.

QuinnRadich (approved):

QuinnRadich:
#sign-off

PR #319, opened 2020-04-18
Return through SetwindowLongPtr

GZPERRA:
The return value must be set with SetwindowLongPtr. The return value of the dialog procedure only informs whether the message was handled.


PRMerger8:
@GZPERRA : Thanks for your contribution! The author(s) have been notified to review your proposed change.

jwmsft (approved):

PR #326, opened 2020-04-23
Update: Describe mpthreat categories table

martyav:
I'm an internal contributor, a writer on @Dansimp's team.

I was asked to update the MPThreats category page:
> Update this topic with descriptions of each MPThreat category

The file was also linted, and one section was converted from a HTML definition list to standard Markdown table syntax. 

I kept in some of the in-line HTML -- span tags -- because they provide hooks for linking to each threat category.

If we should keep the old syntax, I have another branch I could open a PR for.


PRMerger8:
@martyav : Thanks for your contribution! The author(s) have been notified to review your proposed change.

mcleanbyron (approved):

PR #330, opened 2020-04-25
Removed extra word

GordonRudman:
Removed extra 'in'

From "Users are in always in control"
To "Users are always in control"


PRMerger20:
@GordonRudman : Thanks for your contribution! The author(s) have been notified to review your proposed change.

MatchaMatch:

PR #339, opened 2020-04-29
Update dx-graphics-hlsl-function-syntax.md

KarimLUCCIN:
return type doesn't seem to be a factor when trying to disambiguate a function


PRMerger19:
@KarimLUCCIN : Thanks for your contribution! The author(s) have been notified to review your proposed change.

MattSutherlin:
Good, because varying only by return type is just invalid.  However, I'd argue that varying only by function name isn't actually an overload so it's a weird thing to say.

stevewhims (approved):

PRMerger6:

stevewhims:
Thanks, all!

PR #351, opened 2020-05-05
Correct FILEFLAGS and related mask descriptions

riverar:
FILEFLAGSMASK must represent all valid bits, not FILEFLAGS as the docs suggest today. I tried to reduce the wordiness of the descriptions as well. They seemed to be repeating the same general information multiple times.


PRMerger18:
@riverar : Thanks for your contribution! The author(s) have been notified to review your proposed change.

jwmsft (approved):

PR #356, opened 2020-05-06
Make _tmain() return int instead of void.

TedLyngmo:
Some samples won't compile since `_tmain()` is declared `void`.

Signed-off-by: Ted Lyngmo <ted@lyncon.se>


PRMerger9:
@TedLyngmo : Thanks for your contribution! The author(s) have been notified to review your proposed change.

mcleanbyron (approved):

PR #373, opened 2020-05-23
Document more IO_REPARSE_TAGs

saschanaz:
This is the full list of reparse points that are available in `winnt.h`.

They are also documented in https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-fscc/c8e77b37-3909-4fe6-a4ea-2b9d423b1ee4, but some are not included in the header.


PRMerger10:
@saschanaz : Thanks for your contribution! The author(s) have been notified to review your proposed change.

saschanaz:
They are also documented in https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-fscc/c8e77b37-3909-4fe6-a4ea-2b9d423b1ee4

MatchaMatch:

PR #393, opened 2020-06-08
Added Bias-Free Communications Note

Exnav29:
Added a note recognizing that there is wording within this document, which some may feel is exclusionary.  However, this is done only to remain in alignment with the text in the software.


drewbatgit:
#sign-off

PR #407, opened 2020-06-18
Correct error in Return value

bknittel:
The return value of TTM_ENUMTOOLS is always FALSE, whether or not a tool is enumerated. (At least, for TTM_ENUMTOOLSA)


PRMerger12:
@bknittel : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #409, opened 2020-06-18
Very minor update to add TLS to the SSL nomenclature

PeterSmithRedmond:
SSL is the old and familiar term; TLS is the newer and now more correct term. Both are commonly used (still), so we should include both of them on this page. I only updated the first paragraph because (a) it's the most important paragraph (b) we really should include TLS in order to match searches for TLS and (c) it's not actually my technology and I'm not the owner of the page, so I wanted to make as minimal of a change as possible.

### DIFF
--- a/desktop-src/Http/ssl-certificates.md
+++ b/desktop-src/Http/ssl-certificates.md
@@ -8,7 +8,7 @@ ms.date: 05/31/2018
 
 # SSL Certificates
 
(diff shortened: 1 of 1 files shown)


PRMerger12:
@PeterSmithRedmond : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #418, opened 2020-06-25
Remove uniform lane index requirement for Wave

pow2clk:
The HLSL compiler team has agreed to this change, which is more consistent with the later specifications, but failed to update this document. This is consistent with conformant implementations.


PRMerger9:
@pow2clk : Thanks for your contribution! The author(s) have been notified to review your proposed change.

pow2clk:
Mentioning @tex3d so he can see this and weigh in if needed.

PR #420, opened 2020-06-29
It doesn't persist between application restarts

Filoppi:
Not system restarts, like mentioned, which I'm guessing means OS.
When you close and re-open the application, the volume/mute changes are lost with the AUDCLNT_STREAMFLAGS_NOPERSIST, you don't need to restart Windows to lose them.


PRMerger12:
@Filoppi : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Filoppi:
Note that "Application Restart" is the best way of describing because even if we close all the stream, the Windows Mixer will still show that application until the process is actually closed.

drewbatgit (approved):

PR #423, opened 2020-06-30
Fix D3D12 return code page

john-h-k:
The 3 return codes specify don't actually seem to exist (can't find in winerror, magicnumdb, or anywhere), and seem to have just been taken from the equivalent D3D11 errors. I replaced them with the actual `D3D12_ERROR_*` codes i found in winerror/d3d12.h


PRMerger14:
@john-h-k : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #438, opened 2020-07-07
Update values for defined port numbers

erik-anderson:
The values in this documentation do not reflect the actual values in the published header. This change updates it to match the actual values which are actual port numbers (with the exception of the value 0 which maps to the "pick it for me" behavior).


PRMerger13:
@erik-anderson : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (approved):
Thank you!

PR #445, opened 2020-07-10
Replace the flag from int zero to the right enum

siradam7th:
The code that is selected when the variable 'bShaderVisible' is false doesn't compile with the error:
"cannot convert from 'int' to 'D3D12_DESCRIPTOR_HEAP_FLAGS', Conversion to enumeration type requires an explicit cast (static_cast, C-style cast or function-style cast)"
so by using the correct enum D3D12_DESCRIPTOR_HEAP_FLAG_NONE instead of the integer zero, it does not only compile but is also more consistent with usage.


PRMerger15:
@siradam7th : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (approved):
Thank you!

PR #461, opened 2020-07-23
Wide string arguments passed into non wide string function.

wmcnamara:
The arguments passed into MessageBox() are wide strings, and produce an error because MessageBox is not meant for wide strings. This PR fixes this by changing MessageBox() to MessageBoxW().

https://docs.microsoft.com/en-us/windows/win32/learnwin32/example--the-open-dialog-box


PRMerger15:
@wmcnamara : Thanks for your contribution! The author(s) have been notified to review your proposed change.

QuinnRadich (approved):

PR #465, opened 2020-07-28
Windows And Messages: Slight stylistic changes to child, owned window discussion

metathinker:
No technical changes, but these stylistic ones:
- Emphasized more the difference between child and owned windows, and thus between the parent-child and owner-owned relationships of windows.
- Modified the Window Features page table of contents to include the subsections of the child windows section.


PRMerger10:

desktop-src/winmsg/about-windows.md
```diff
@@ -153,9 +153,11 @@ An application can size a window so that it is extremely large; however, it shou
...
 
-A window can own, or be owned by, another window. An owned window always appears in front of its owner window, is hidden when its owner window is minimized, and is destroyed when its owner window is destroyed. For more information, see [Owned Windows](window-features.md).
+A window that has no parent, or whose parent is the desktop window, is called a *top-level window*. An application can use the [**EnumWindows**](/windows/win32/api/winuser/nf-winuser-enumwindows) function to obtain a handle to each top-level window on the screen. **EnumWindows** passes the handle to each top-level window, in turn, to an application-defined callback function, [**EnumWindowsProc**](https://msdn.microsoft.com/library/ms633498(v=VS.85).aspx).
+
+A top-level window can own, or be owned by, another window. An *owned window* always appears in front of its owner window, is hidden when its owner window is minimized, and is destroyed when its owner window is destroyed. For more information, see [Owned Windows](window-features.md#owned-windows).
```
hickeys, line 160: "A top-level window can own..." heavily implies that child windows can't have children of their own. Is that the case?

hickeys (commented):
Just have the one clarifying question. Otherwise everything looks great.

desktop-src/winmsg/about-windows.md
metathinker: Child windows can have children of their own. However, "window 2 is owned by window 1" is a separate relationship from "window 2 is a child of window 1". What I was trying to say here is "if window 1 is the parent or a child of some window 2, then window 2 cannot own or be owned by any window 3". In other words, a window can be in at most one of the 2 relationships "parent-child" or "owner-owned".

Adding @chrisGuzak who has corrected me at least once in the past when I imprecisely spoke about owner and parent windows as the same thing. Chris, perhaps you may have a better way to explain the distinction?

ChrisGuzak: this looks good to me. thanks for the edit.

ChrisGuzak (approved):

PRMerger9:
@hickeys : Thanks for your contribution! The author(s) have been notified to review your proposed change.

hickeys:
#sign-off

PR #482, opened 2020-08-17
Replace dl-elements by simple level-3-headings

ElmarJ:
There was no way to place a small comment, so making this pull request for a small change.

The sections about different Windows versions are created using dl, dt and dd HTML tags. Not only does that seem semantically incorrect (because this is not a dictionary with descriptions / definitions), but more importantly: they currently aren't formatted as a header on the microsoft docs site (see https://docs.microsoft.com/nl-nl/windows/win32/sysinfo/acquiring-high-resolution-time-stamps). This makes the structure less obvious and makes reading a bit harder than necessary. This change fixes that by using simple headers.

Thanks!


PRMerger14:
@ElmarJ : Thanks for your contribution! The author(s) have been notified to review your proposed change.

mcleanbyron (approved):

PR #488, opened 2020-08-19
Update using-the-compression-api-in-block-mode.md

walbourn:
This code sample won't build for x64 native. I updated it to work for both x86 & x64

> Note that the decompressor doesn't handle 32-bit overflow so that's probably something a security person should review.


PRMerger8:
@walbourn : Thanks for your contribution! The author(s) have been notified to review your proposed change.

mcleanbyron (approved):

PR #489, opened 2020-08-19
Fix broken Note rendering in -gdiplus-types-of-coordinate-systems-about.md

ericmutta:
The note section with the words `If you don't specify a pen width ...` doesn't seem to render correctly when converted to HTML (see live page here https://docs.microsoft.com/en-us/windows/win32/gdiplus/-gdiplus-types-of-coordinate-systems-about)

I suspect this may have to do with the blank line just before the code example that calls the Pen constructor. I have tried using two "break" tags to keep the blank line in a way that will cause the resulting HTML to properly render the Note (i.e with highlights to call attention to it). **This may not be the correct fix so I welcome further edits to this PR.**

The broken note renders as follows:

gh_note_bad_render

...whereas it should render similar to this note (from here https://docs.microsoft.com/en-us/windows/win32/gdiplus/-gdiplus-types-of-bitmaps-about):

gh_note_good_render


PRMerger13:
@ericmutta : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #490, opened 2020-08-20
Remove string "1000" from -gdiplus-global-and-local-transformations-about.md

ericmutta:
Remove instances of the string "1000" from several locations where it does not seem to relate to the surrounding content.

Also note that the image shown below the sentence `The following illustration shows the matrices involved in the transformation.` is incorrect (seems to be a copy-and-paste error and I am not sure how to link to the right image so I mention that here to bring the error to the PR reviewer's attention).


PRMerger15:
@ericmutta : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger9:

stevewhims:
Thanks, @ericmutta. It looks like the illustration showing the matrices involved in the transformation was either never generated or has been deleted. So I removed the text and the duplicate link.

PR #494, opened 2020-08-21
Update signtool.md for /verify option

rackley096794:
Update to clarify signtool.exe verify option checks *only the EMBEDDED* signature if no catalog options are specified on the command line. Code review of the tool (/ds/security/cryptoapi/pkisign/tools/signtool/*) shows this is correct. However, this is not clear from any of our docs, nor does the tool output indicate that the signature is embedded if it is an embedded signature.

### DIFF
--- a/desktop-src/SecCrypto/signtool.md
+++ b/desktop-src/SecCrypto/signtool.md
@@ -331,7 +331,9 @@ The following display options apply to all SignTool commands.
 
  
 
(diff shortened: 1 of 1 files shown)


PRMerger14:
@rackley096794 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

hkhmsft:
Thanks for adding this detail, @rackley096794 ! Changes should be live soon.

PR #495, opened 2020-08-24
Fix error with longPathAware example

GenericMadScientist:
In windowsSettings, the new namespace was called ws3 but then ws2 was used on the next line.


PRMerger18:
@GenericMadScientist : Thanks for your contribution! The author(s) have been notified to review your proposed change.

mcleanbyron (approved):

PR #497, opened 2020-08-28
Fix memory leak in xinput-and-directinput.md

SamB:
At least, I hope I did this right.

I just happened across a report of the problem at http://www.virtualdub.org/blog/pivot/entry.php?id=386 and figured I might as well see if the relevant page had made it to GitHub or not.


PRMerger15:
@SamB : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Karl-Bridge-Microsoft (approved):
Looks good

PR #498, opened 2020-08-29
Fix images with spaces in their filenames

vegard:
Three figures were not showing up correctly because the filenames had spaces in them. Escaping the spaces using HTML encoding (%20) fixes the problem.


PRMerger14:
@vegard : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #512, opened 2020-09-12
Update security-properties.md

ryannewington:
The documentation incorrectly states that `I` is supported on it's own, while `S` is not. The reverse is actually true. 

The following documentation shows that `S` is supported on it's own, while `I` is not.

https://docs.microsoft.com/en-us/windows/security/identity-protection/smart-cards/smart-card-certificate-requirements-and-enumeration#kdc-certificate


PRMerger9:
@ryannewington : Thanks for your contribution! The author(s) have been notified to review your proposed change.

MatchaMatch (approved):

MatchaMatch:

PR #516, opened 2020-09-12
Update direct3d-11-advanced-stages-tessellation.md

mcleary:
The names in "Types of Partitioning" should be all in lower case as the compiler expects it in that way.


PRMerger7:
@mcleary : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (approved):
Thank you!

PR #541, opened 2020-09-23
Add Multicast Programming Sample

aphage:


PRMerger18:
@aphage : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims:
Hi, @aphage, and thanks for this contribution. I've asked an SME to give it a review, so we'll wait for that before completing the PR.

Thanks again!
-Steve

nigriMSFT:
This sample seems very specific to a certain use case (which is unclear since there is no PR description or high level code comments). I would hesitate to add this to an existing msdn page that simply describes how to join and leave multicast groups.

@aphage can you at least state what is being accomplished in the sample code?

If the usage is not general enough to add to msdn pages, maybe it would be better suited in the WinSock SDK samples repo
https://github.com/microsoft/Windows-classic-samples/tree/master/Samples/Win7Samples/netds/winsock

aphage:
"consumer" traverse the network card to join the multicast address 225.233.233.233 monitor port 7777, receive data

"producer" traverses the network card and sends a string message of "Alice foo foo !" to the multicast address 225.233.233.233 port 7777

I think there are 3 points to note

1. Socket bind, Windows cannot bind multicast addresses, such as bind(225.233.233.233), Linux can work, Windows needs to bind like ordinary sockets, such as bind(0.0.0.0)

2. When there are multiple network cards, if the IP address of the network card is not specified "mreq.imr_interface.s_addr = htonl(INADDR_ANY); ", one of the network cards is selected by default, so it may happen that multicast messages cannot be received or not sent out

3. When using IP_ADD_SOURCE_MEMBERSHIP, imr_sourceaddr cannot be INADDR_ANY, at least for Windows

nigriMSFT:
From looking at the code it looks like the "consumer" portion is joining a particular multicast group on all interfaces of the machine and doing a receive loop, while the "producer" portion is sending out a message destined for the same multicast group from each interface of the machine. These might be reasonable examples that others can benefit from. However, I would highly encourage you to open a PR in the Windows API samples repo as I think it would be a better fit. 
https://github.com/microsoft/Windows-classic-samples/tree/master/Samples/Win7Samples/netds/winsock

stevewhims:
What do you think, @aphage? It'd be great if you could contribute to a samples repo, as @nigriMSFT advises.

Are you ok if I go ahead and close this pull-request?

Thanks!
Steve

PR #560, opened 2020-10-11
Give an example where processor groups are not filled up.

jaykrell:
I am not an expert and do not know if this is true.
It folllows from the earlier statement, but then an example is given that does not demonstrate the point, which I found confusing.


PRMerger20:
@jaykrell : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Karl-Bridge-Microsoft:
Hi @jaykrell - thanks for your suggestion. 

From the preceding section:
Support for systems that have more than 64 logical processors is based on the concept of a processor group, which is a static set of up to 64 logical processors that is treated as a single scheduling entity. Processor groups are numbered starting with 0. Systems with fewer than 64 logical processors always have a single group, Group 0.

So the example as stated makes sense:
For example, a system with 128 logical processors would have two processor groups with 64 processors in each group, not four groups with 32 logical processors in each group.

PR #561, opened 2020-10-11
Fix IPv6 format typo in isInNetEx example

theaquamarine:
IPv6 subnets are formatted with `::` in the prefix, as used in the description at the top - add the missing colons to an example.


PRMerger9:
@theaquamarine : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #566, opened 2020-10-14
detail halfway cases in round documentation

Jamather:
This rounds halfway cases to the nearest even.  We should be explicit about this especially because it differs from C++'s std::round which rounds halfway cases away from zero.


PRMerger7:
@range4life : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (approved):
Thank you!

PR #567, opened 2020-10-16
connect call was failing due to the value of "hints.ai_family" paramt…

GopalSir:
…er of "hints"  variable of type addrinfo being set to "AF_UNSPEC"  .

Added a warning in note section to change AF_UNSPEC to specific version of IP ( AF_INET or AF_INET6 ) in case the connect call fails .


PRMerger14:
@GopalSir : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (approved):
Thank you!

PR #573, opened 2020-10-29
Abstract message meaning from cause and clarify causes.

definedrisk:
Meaning should be "Notifies a window that the user desires a context menu to appear."

The current document states "Notifies a window that the user clicked the right mouse button" which is 
 sent via messages such as WM_RBUTTONDOWN.


PRMerger7:
@definedrisk : Thanks for your contribution! The author(s) have been notified to review your proposed change.

jwmsft (approved):

PR #609, opened 2020-11-09
Update refreshing-a-dynamic-object.md

HerbertMauerer:
the update to entryTTL has be replicate before the object expires. e.g. for groups it will end up with an empty membershp list when you resurrect a group from expired state.

Filed as BUG 29608897 as end-status is a group with no members on one replica, and members on the other.


PRMerger12:
@HerbertMauerer : Thanks for your contribution! The author(s) have been notified to review your proposed change.

MatchaMatch:

PR #616, opened 2020-11-10
Fix links at bottom of page.

markjulmar:
HTML tags cause the rendering engine to not generate anchor tags. Removing them fixes the issue.


markjulmar:
#signoff

mcleanbyron (approved):

PR #631, opened 2020-11-18
Update disabling-terminal-services-features.md

kheldorn:
Undocumented registry key that prevents the client from handing the local drives to the server for mapping. While the keys for Clipboard and Printer will grey-out the options in mstsc.exe the Drive key will not. Drives will not get mapped with DisableDriveRedirection set to 1 though.


PRMerger18:
@kheldorn : Thanks for your contribution! The author(s) have been notified to review your proposed change.

mcleanbyron:
Thanks for submitting this pull request. I am following up with the engineering team that owns this feature area to confirm whether this key should be documented.

mcleanbyron (approved):

PR #635, opened 2020-11-19
Update so-keepalive.md

maolson-msft:
The current text causes confusion about whether keepalives are enabled or disabled by default, due to a statement along the lines of "by default the timeout is 2 hours and interval is 1 second" [these settings are only relevant when keepalives are actually enabled]. This statement is actually redundant because the default settings are mentioned a little earlier in the text, so the fix is to delete this confusing statement. I also fixed some related poor wording.


PRMerger9:
@maolson-msft : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (approved):

PR #636, opened 2020-11-19
Update sio-keepalive-vals.md

antonfirsov:
Suggestion similar to #635. Only difference that I'm still mentioning the defaults, since that info is not redundant in sio-keepalive-vals.md

/cc @maolson-msft


PRMerger9:
@antonfirsov : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (approved):
Thank you!

PR #640, opened 2020-11-24
Update code example to not use undefined variable.

Shipleaves:
The example code declares a variable `HINST hinst` but later uses an undefined variable `hinstance`. Presumably this should have been the previously defined `hinst` variable.


PRMerger7:
@Shipleaves : Thanks for your contribution! The author(s) have been notified to review your proposed change.

hickeys:
Good catch, thank you!

PR #642, opened 2020-11-25
Fix old shared_ptr and broken _aligned_malloc links

pyjarrett:
- Updated to (hopefully) language-agnostic `https://docs.microsoft.com/cpp/...`, tested with en-us, ru-ru and es-es.  I'm not sure how your language redirects work, but just removing the locale from the url seems to work based on changing my profile's language.
- I'm also not sure if you want to cross-link across your docs, or if you have a better internal way of doing this.


stevewhims (approved):

stevewhims:
Hi, @pyjarrett. Yes, your instinct is correct about dropping the language code (locale) from the url. If you've got a link from one `https://docs.microsoft.com` topic to another, then you can also drop that part of the url if you like (the benefit being that they will work in the Microsoft Help Viewer as well as online). Resulting in, for example [**_aligned_malloc**](/cpp/c-runtime-library/reference/aligned-malloc). You need the leading '/' there. You don't need to escape underscores if they're within pairs of ** (or similar).

Thanks!
Steve

PR #651, opened 2020-12-08
Powershell example should have -ne instead

Ryaed:
Powershell example should have -ne instead of -notmatch.  UserFriendlyName is a Unit16 array and the idea i believe is to ignore items in the array with a value of 0. NotMatch is a regex lookup at the string that will exclude important non-zero values that contain a zero like 105  which is the Ascii code for uppercase "E"

https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_comparison_operators?view=powershell-7.1#-notmatch


PRMerger10:
@Ryaed : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (approved):

PR #675, opened 2020-12-22
Update using-thread-local-storage.md

huzaifa-d:
Changing code to be out-of-the box compatible with C++ 14.


PRMerger20:
@theShaan : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (approved):

stevewhims:
Thanks for the PR, @theShaan. I've changed the macro from LPSTR (which is not const) to LPCSTR (which is const). That achieves the same good ends as your change does, but it also stays in the same spirit as the original code (which was to use the Windows type macros).

PR #684, opened 2020-12-28
Update dx-graphics-hlsl-to-load.md

LuridHound:
Misleading description of input parameters.
For example in the description of the function "GetDimensions"
we have a template type of an input variable, that can be changed amid parameters.
https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-to-getdimensions
But in the current function ( I am speaking about "Load") is said, that **each** parameter has type int.
And when you scroll down the page, you can observe, that objects of different types also accept
different types of parameters

### DIFF
--- a/desktop-src/direct3dhlsl/dx-graphics-hlsl-to-load.md
+++ b/desktop-src/direct3dhlsl/dx-graphics-hlsl-to-load.md
@@ -28,7 +28,7 @@ Reads texel data without any filtering or sampling.
 </tbody>
 </table>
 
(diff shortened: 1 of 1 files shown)


PRMerger9:
@LuridHound : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger13:
@LuridHound : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (approved):

PR #692, opened 2021-01-06
[D3D12] Update Common state promotion examples

Lectem:
This changes the Common state promotion example to explain that promotion from a read promoted state to another is implicit, but that it is not the case for write states.


PRMerger18:
@Lectem : Thanks for your contribution! The author(s) have been notified to review your proposed change.

jenatali (approved):
Looks good to me, a much more in-depth edit than I was expecting but I like the extra examples.

stevewhims (approved):

PR #695, opened 2021-01-07
Update anonymous-pipe-operations.md

cminyard:
It appears that overlapped operations work on anonymous pipes, and it makes not sense that they don't.  Is this true?  Overlapped operations on asynchronous pipes seems to work fine for me on Windows 10.


PRMerger9:
@cminyard : Thanks for your contribution! The author(s) have been notified to review your proposed change.

cminyard:
Well, it appears that the documentation is right and ti doesn't work. Bummer :(.

PR #694, opened 2021-01-07
Update creating-a-child-process-with-redirected-input-and-output.md

cminyard:
The third parameter to SetHandleInformation() value, which needs to be HANDLE_FLAG_INHERIT in this case.

PR #696, opened 2021-01-08
Revert "Update creating-a-child-process-with-redirected-input-and-output.md"

cminyard:
Reverts MicrosoftDocs/win32#694

I was wrong about this, I'm sorry.  It seemed to fix my problem, but I got the In/Out Rd/Wr messed up.  This is disabling inheritance on the local side of the pipe, which is exactly what you want to do.  So this should be reverted.


PRMerger8:
@cminyard : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Karl-Bridge-Microsoft (approved):
Ah you're right. I missed a qualification on the https://docs.microsoft.com/en-us/windows/win32/api/handleapi/nf-handleapi-sethandleinformation topic that 0 is a valid value.

PR #701, opened 2021-01-16
Added .reg file code to enable Long Paths

iSatishYadav:
Added .reg file code to enable Long Paths so that people can create a .reg file and avoid accidental changes in Registry Editor.


PRMerger19:
@iSatishYadav : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #713, opened 2021-01-23
Cleanup the Shared Registry Keys page; use language inclusive of ARM architectures

franklesniak:
The "Shared Registry Keys" page included a table of registry key locations that was poorly formatted for markdown and very difficult to understand unless you viewed the raw .md page. I cleaned up the table and listed the full registry key paths to remove any ambiguity.

The page was also written exclusively for 32-bit Intel IA-32 (x86) applications running on 64-bit AMD64 / Intel x86-x64 Windows installations, ignoring ARM and ARM64 installations of Windows and their corresponding behavior. So, I updated the language to be more inclusive, while retaining an example of the common x86 on AMD64 scenario for clarity.

Lastly, the information pertaining to HKCU was confusing, so I cleaned it up.


PRMerger13:
@franklesniak : Thanks for your contribution! The author(s) have been notified to review your proposed change.

franklesniak:
It's not my text and shouldn't affect this PR, but it occurred to me that the HKEY_LOCAL_MACHINE\SOFTWARE\Classes\CLSID case for Windows Server 2008, Windows Vista, Windows Server 2003, and Windows XP is a little ambiguous.

Does it mean that **both** redirection and reflection occur only for CLSIDs that do not specify InprocServer32 or InprocHandler32?

-or-

Does it mean that redirection always occurs, but that reflection only occurs for CLSIDs that do not specify InprocServer32 or InprocHandler32?

QuinnRadich (approved):

QuinnRadich:
Frank.

Thanks for submitting this PR, it seems like a good tune-up to an out-of-date page.

As for your later comment about redirection, I'll investigate this further. Feel free to open an Issue about it if you see fit.

Many thanks!

QuinnRadich:
#sign-off

PR #719, opened 2021-01-29
Update sfvm-fsnotify.md

mihayash:
Hi experts, this is Mie Hayashi from Developer support CSS Japan. 
while the investigation for support case, we found a doc bug.
Please see the attachment for details.

Sorry, I have deleted the attachment. I will mail details.

### DIFF
--- a/desktop-src/shell/sfvm-fsnotify.md
+++ b/desktop-src/shell/sfvm-fsnotify.md
@@ -24,7 +24,7 @@ Notifies the callback object that an event has taken place that affects one of i
 ```C++
 SFVM_FSNOTIFY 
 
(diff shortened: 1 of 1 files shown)


PRMerger10:
@mihayash : Thanks for your contribution! The author(s) have been notified to review your proposed change.

mcleanbyron (approved):

ChrisGuzak (commented):
this is a step in the right direction so I'll approve but...

This describes how to **call** the IShellFolderViewCB::MessageSFVCB but 3rd parties should almost never do that (testing is the exception), the system makes these calls and does not need this documentation. 

Instead it should show how to implement this method and cast the wParam and lParam values.

ideally this would not use C style casts, use C++ casts only.

PR #726, opened 2021-02-03
Update Protocol Versions for Cipher Suites

poloplayerrl:
When TLS 1.1 and TLS 1.2 support was added to Vista / Server 2008 SP2 we did not update the documentation that specifies which protocol versions can be used for each cipher suite.


PRMerger20:
@poloplayerrl : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #727, opened 2021-02-04
Updated Protocol Versions For Cipher Suites

msjuburke:
When TLS 1.1 and 1.2 support was added for Vista/Server 2008 SP2 we didn't update the documentation that specifies which cipher suites can use the additional protocols.


PRMerger9:
@msjuburke : Thanks for your contribution! The author(s) have been notified to review your proposed change.

hkhmsft:
Thanks for the update, @msjuburke !

PR #732, opened 2021-02-06
Note maximum number of tags that can be stored w/ System.Keywords

mguinness:
See [Maximum vector size for properties in Windows Search](https://stackoverflow.com/questions/52729984) for a hard limit on tags that can be stored.

### DIFF
--- a/desktop-src/properties/props-system-keywords.md
+++ b/desktop-src/properties/props-system-keywords.md
@@ -27,7 +27,7 @@ propertyDescription
 
 ## Remarks
 
-PKEY values are defined in Propkey.h.
+PKEY values are defined in Propkey.h.  A maximum of 20 keywords can be stored.
 
 ## Related topics
 
--- a/desktop-src/properties/props-system-keywords.md
+++ b/desktop-src/properties/props-system-keywords.md
@@ -27,7 +27,7 @@ propertyDescription
 
 ## Remarks
 
-PKEY values are defined in Propkey.h.  A maximum of 20 keywords can be stored.
+PKEY values are defined in Propkey.h. A maximum of 20 keywords can be stored.
 
 ## Related topics
 


PRMerger13:
@mguinness : Thanks for your contribution! The author(s) have been notified to review your proposed change.

mguinness:
I did some further research and found the [Windows Search Protocol](https://winprotocoldoc.blob.core.windows.net/productionwindowsarchives/MS-WSP/[MS-WSP].pdf) document.  Below is an extract where maximum size for System.Keywords is listed as 512.

| Property Name | GUID | propID | inInverted Index | isColumn | columnIndexType | type | MaxSize | VectorProperty | Description |
|-|-|-|-|-|-|-|-|-|-|
| System.Keywords | {F29F85E0-4FF9-1068-AB91-08002B27B3D9} | 5 | TRUE | TRUE |  | String | 512 | TRUE | The keywords for the item. Also referred to as tags. |

GrantMeStrength:
Researching

KalleOlaviNiemitalo:
\[MS-WSP\] v20210625 §2.2.5 says MaxSize is the maximum size **in bytes** -- not the maximum number of keywords.  I don't know how many bytes of overhead it adds for each keyword in the vector, in addition to the bytes needed for the characters themselves.

mguinness:
Thanks, I found the Standard Properties section you mentioned in [Windows Search Protocol](https://winprotocoldoc.blob.core.windows.net/productionwindowsarchives/MS-WSP/%5BMS-WSP%5D-210625-diff.pdf) document.

> **MaxSize**: An unsigned integer indicating the maximum size, in bytes, that this property can have stored or retrieved by the Generic Search Service (GSS).

I tried varying character lengths for the tags and the limit still appears to be 20 tags, so that seems hardcoded rather than a byte size limitation.

mguinness:
A year out from the last comment I thought I'd bump issue to see if the size limit can be verified and documented.

GrantMeStrength:
ugh. let me blow the dust off it.

GrantMeStrength (approved):

GrantMeStrength:
Closing due to age (4+ years). The content may have changed since this was filed. If this is still relevant, please feel free to open a fresh PR. Thanks for the contribution!

mguinness:
Is it really worth submitting another PR, grant me strength indeed.

PR #733, opened 2021-02-07
Update enumerating-all-processes.md

aloofbit:
Hi, I was trying to use this code and most of my applications showed "<unknown>". I was wondering if we could add a note about x64/x64 mode applications. I'm really new to trying to use C++ so I am not sure if this is something that was supposed to be very obvious. Please let me know if there is a better way to do this!

### DIFF
--- a/desktop-src/psapi/enumerating-all-processes.md
+++ b/desktop-src/psapi/enumerating-all-processes.md
@@ -8,8 +8,10 @@ ms.date: 05/31/2018
 
 # Enumerating All Processes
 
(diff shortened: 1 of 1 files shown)


PRMerger13:
@aloofbit : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger12:
@Karl-Bridge-Microsoft : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Karl-Bridge-Microsoft (approved):

PR #737, opened 2021-02-10
Add link from 'Raytracing HLSL Reference' page to State Objects page

clandrew:
There's this hub page linking to all the places you can find DXR HLSL syntax. 
In the interest of completeness, hub page should also link to 'State objects', since you can also specify DXR stuff through those.
It's formatted as a link rather than a row in the table because the linked page is not technically part of this section.


PRMerger6:
@clandrew : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (approved):

PRMerger15:

clandrew:
Thank you @stevewhims for taking a look!

PR #742, opened 2021-02-12
Fix documentation for HLSL Gather*

### LINKED ISSUE #2251: Improve HLSL docs describing gather operations
https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/texture2d-gather
Needs a better description than "Samples a Texture2D and returns all four components."
Same for the various overloads

https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-to-gather
should probably have a function prototype.

clandrew:
The API reference docs for all the Gather* aren't correct and appear like they might've been erroneously copypasted from Sample. After all it isn't really right to say that Gather does a sample (Sample is what does a sample), it sort of does the first part of a very limited kind of sample. It would be better to say more clearly what it does.

This issue affects the whole family of methods (cmp / single-channel / overloads that return status) so it touches a lot of places to fix all of them.

An example of an affected page is https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/texture2d-gather

This is intended to fix https://github.com/MicrosoftDocs/feedback/issues/2251

Additionally, there was a spot where Texture2DMSArray::GetSamplePosition also had the same text "Samples a texture and returns all four components" which is just clearly wrong so this fixes it.

### DIFF
--- a/desktop-src/direct3dhlsl/sm5-object-texture2d-gather.md
+++ b/desktop-src/direct3dhlsl/sm5-object-texture2d-gather.md
@@ -1,6 +1,6 @@
 ---
 title: Texture2D::Gather(S,float,int) function
-description: Samples a texture and returns all four components.
+description: Returns the four texel values that would be used in a bi-linear filtering operation.
 ms.assetid: 5d196c1c-8cc9-4add-9d33-654294314ee2
 keywords:
 - Gather function HLSL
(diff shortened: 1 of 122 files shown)


clandrew:
Cleaned up unnecessary commit from commit history

clandrew:
@stevewhims Would you be able to review this sometime or point me to someone who can? (No hurry) I tried to add you as a reviewer but it doesn't appear I have permission to

stevewhims (approved):
Looks good, thanks for fixing these, Claire!

clandrew:
> Looks good, thanks for fixing these, Claire!

no problem!

PR #747, opened 2021-02-14
WM_COPY must return a nonzero value on success.

code-creator:
If you overwrite WM_COPY, and return zero then an edit control will not delete the Text in response to a WM_CUT message.
I think it does internally call the WM_COPY handler, and then deletes the text only if WM_COPY handler returned a nonzero value.
Please review/correct the documentation.


PRMerger18:
@code-creator : Thanks for your contribution! The author(s) have been notified to review your proposed change.

MatchaMatch (approved):

PR #759, opened 2021-02-19
Fix wParam value

rtanabe999:
If want to search from the beginning of the list box, should be set wParam to -1, not 1.


PRMerger17:
@rtanabe999 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

MatchaMatch:

PR #764, opened 2021-02-22
Update cb-findstringexact.md

ryedigaryan:
Seems that wParam should be -1 to search the entire list box from the beginning.
The message result is the same for both 1 and -1 because when the search reaches the bottom of the list box, it continues from the top of the list box back to the 1st item.
It is -1 for [CB_FINDSTRING](https://github.com/MicrosoftDocs/win32/blob/docs/desktop-src/Controls/cb-findstring.md
) and [CB_SELECTSTRING](https://github.com/MicrosoftDocs/win32/blob/docs/desktop-src/Controls/cb-selectstring.md).

### DIFF
--- a/desktop-src/Controls/cb-findstringexact.md
+++ b/desktop-src/Controls/cb-findstringexact.md
@@ -27,7 +27,7 @@ Finds the first list box string in a combo box that matches the string specified
 *wParam* 
 </dt> <dd>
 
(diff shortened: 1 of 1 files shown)


PRMerger9:
@ryedigaryan : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #769, opened 2021-02-25
Remove frquery.h from HLSL vector doc

pow2clk:
No idea why this was here. It has nothing to do with HLSL

### DIFF
--- a/desktop-src/direct3dhlsl/dx-graphics-hlsl-vector.md
+++ b/desktop-src/direct3dhlsl/dx-graphics-hlsl-vector.md
@@ -1,5 +1,5 @@
 ---
-title: Vector Type (Frquery.h)
+title: Vector Type (HLSL)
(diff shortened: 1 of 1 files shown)


PRMerger8:
@pow2clk : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (changes requested):
We don't want to remove metadata, nor the Requirements block. Where the wrong header file is being cited, we should correct that to the right one.

pow2clk:
There is no header, it's an HLSL intrinsic. It's available when using the HLSL compiler without any header.

stevewhims (approved):
I see; it's a conceptual topic, so removing these metadata fields is ok, Approved!

PRMerger14:
@pow2clk : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #775, opened 2021-03-04
Fix case of XML attribute in taskschedulerschema-exec-actiongroup-element.md

skissane:
On the Exec element, the correct case of the id attribute is `id` not `Id`.

You can confirm this by checking the schema, or trying to create a task from XML using `schtasks.exe`

XML attribute names are case-sensitive, and using the wrong case causes an error.


PRMerger13:
@skissane : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (approved):

PR #776, opened 2021-03-04
Default return values are wrong, the minus got lost

code-creator:
Same prob as in WM_VKEYTOITEM where I had created a pull request which was accepted.
Please change here too.

### DIFF
--- a/desktop-src/Controls/wm-chartoitem.md
+++ b/desktop-src/Controls/wm-chartoitem.md
@@ -50,11 +50,11 @@ Handle to the list box.
 
 ## Return value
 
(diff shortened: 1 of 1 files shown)


code-creator:
Seems somehow in the process of bringing the documentation to github, negative numbers lost their "-".
See my (merged) pull request on WM_VKEYTOITEM, and
https://github.com/MicrosoftDocs/win32/pull/764 where it was the same problem.
Sentences with "... returns -1" somehow in GitHub transformed to "... returns  1", for example. Instead of the minus sign always was an extra space before the number.
I suspect there are other places in the documentation too, where the "-" is lost. 
Perhaps you can automatically check them against older documentation, please.

GrantMeStrength:
Thank you for catching this and submitting the fix!

-John

PR #782, opened 2021-03-09
Correct description of KF_REPEAT

tjs137:
KF_REPEAT does not select the key repeat count (bits 0-15), but rather the previous key-state flag (bit 30).


PRMerger20:
@tjs137 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Karl-Bridge-Microsoft (approved):

PR #799, opened 2021-03-17
Standard subject to change warning for RtlGetUnloadEventTraceEx

henke37:
The non Ex version has the warning, so I assume that it should be on the Ex version too.

### DIFF
--- a/desktop-src/DevNotes/rtlgetunloadeventtraceex.md
+++ b/desktop-src/DevNotes/rtlgetunloadeventtraceex.md
@@ -17,6 +17,8 @@ api_location:
 
 # RtlGetUnloadEventTraceEx function
 
+\[This function may be changed or removed from Windows without further notice.\]
+
 Retrieves the size and location of the dynamically unloaded module list for the current process.
 
 ## Syntax


PRMerger10:
@henke37 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

mcleanbyron:
Thanks for submitting this change request against our docs. Where did this information come from? Were you told that this function may be removed in a future release by a Microsoft representative or employee?

henke37:
As I wrote, this was a logical inference. This function is an Ex version of a function that does have this same warning and I assumed that this version should have it too. But I do not have an authoritative claim that this is indeed the case. I was hoping that the PR review process would locate one.

mcleanbyron:
Thanks for the clarification! Sorry, I missed your first comment. 

Releasing an "Ex" version of a function does not necessarily mean that any restrictions/etc. for the earlier version also apply to the "Ex" version, and as far as I can tell from a quick look at the source code there is no indication that this statement applies to RtlGetUnloadEventTraceEx. Given this, I'm closing this pull request.

PR #807, opened 2021-03-31
Fix code example to assign to hr before reading it

praeclarum:
The example read the variable `hr` before it was checked. I fixed the code to handle the possible error from the function call.


PRMerger8:
@praeclarum : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #825, opened 2021-04-14
Fixed position of image and explanation text

Ohjurot:
Hello,

The changed text and image was explaining how the example of  "Root Signature Version 1.0" works and looks like, but was placed under "Root Signature Version 1.1". I have only move it to the top. No changes to the text itself.


PRMerger7:
@Ohjurot : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #831, opened 2021-04-22
Update countbits.md

LuridHound:
https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/countbits---sm5---asm-
Here we have the following description :

>    **_Counts the number of bits set in a number_**

Meantime in the current page of documentation the following :

>    **_Counts the number of bits (per component) in the input integer_**

And then the question appears. 
  

> Does it work like **_sizeof_** in C++, or it work like the function from the first link?

In attempt to avoid this question, added one word in description.

### DIFF
--- a/desktop-src/direct3dhlsl/countbits.md
+++ b/desktop-src/direct3dhlsl/countbits.md
@@ -17,7 +17,7 @@ api_location:
 
 # countbits function
 
-Counts the number of bits (per component) in the input integer.
+Counts the number of bits (per component) set in the input integer.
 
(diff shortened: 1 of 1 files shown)


PRMerger7:
@LuridHound : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger14:
@LuridHound : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (approved):

PR #849, opened 2021-05-13
Fix wonky escaping of table titles, update pub date

ericlaw1979:
Note: Please review carefully-- I haven't worked in these docs previously and I don't know if there was some reason the `<dl><dt><dd>` HTML markup was embedded here, but it was causing malformation of the markdown titles:

image


PRMerger12:
@ericlaw1979 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #855, opened 2021-05-18
HLSL compiler #pragma message fix

fdwr:
The example doesn't actually work as-is (nothing is printed by fxc.exe), but just wrapping it with parens fixes it and prints the message.


PRMerger20:

stevewhims (approved):

PR #860, opened 2021-05-28
Update international-components-for-unicode--icu.md: CoInitializeEx is only needed on Windows < 1903

metathinker:
[That's what Raymond Chen says, anyway.](https://devblogs.microsoft.com/oldnewthing/20210527-00/?p=105255) @jefgen or @daniel-ju - I believe you're the domain experts here; would you kindly fact check me when you get time?

### DIFF
--- a/desktop-src/Intl/international-components-for-unicode--icu-.md
+++ b/desktop-src/Intl/international-components-for-unicode--icu-.md
@@ -96,7 +96,7 @@ Then, you can call whatever ICU C API from these libraries you want. (No C++ API
 > [!Note]  
 >
 > - This is the configuration for “All Platforms”.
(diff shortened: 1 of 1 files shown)


PRMerger14:

daniel-ju (approved):
Thank you for updating the doc! The change looks good/accurate to me.

Karl-Bridge-Microsoft (approved):

jefgen:
@metathinker Thanks for updating this! And thanks @daniel-ju for reviewing this!

Yes, this is correct. As of 19H1 (aka version 1903) Win32 apps don't need to call `Co­Initialize­Ex` anymore before using ICU, so this change looks good to me as well. :)

PR #861, opened 2021-05-28
I think the "title" in the article was meant to be replaced with Jeky…

danielkrajnik:
Replaced all "title" with "{{ page.title }}"


PRMerger10:
@danielkrajnik : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (changes requested):
So, here the word "title" actually means "game". As in "The game studio managed to release two triple-A titles this year." Anyway, the change in this PR is not what we want. I'd suggest either closing the PR, or else changing each instance of "title" (when used in the context I describe) to something like "game or app".

PR #866, opened 2021-06-01
Fix typo in What's New in DirectWrite

PeterCon:


PRMerger8:
@PeterCon : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (approved):
Thank you, and approved. However (just a low-pri though here), I do feel that "No X nor Y" is valid, and here means "Zero features, and zero APIs".

Whereas "No X or Y" means "Zero features or APIs" (which could be read as "Zero features or some indeterminate number of APIs"). So that seems to have the same ambiguity (albeit a very minor one) that "Neither X or Y" does (which is always clearer with "nor"). Also note that we still have "added nor updated" a few words later, so there's a slight inconsistency with the sentence's adherence to strict use of "nor" vs. "or".

I'd have left this as-is myself, but I also feel that this change is fine. :)

PR #884, opened 2021-06-08
 Fix the endless loop caused by the'%%num' parameter

MountCloud:
Fix the endless loop caused by the'%%num' parameter.


PRMerger16:
@MountCloud : Thanks for your contribution! The author(s) have been notified to review your proposed change.

MountCloud:
The parameter in the format of '%%num' was extracted from the windows10 security event.It leads to endless loops.

Karl-Bridge-Microsoft (approved):

Karl-Bridge-Microsoft:
Hi @MountCloud - thanks for your suggestion.

PR #886, opened 2021-06-08
Fixed CODECAPI_AVLowLatencyMode variant type

adepke:
Using the suggested `VT_BOOL` variant type throws an exception with the message `WinRT originate error - 0x80070057 : 'Invalid argument (VT_UI4 != pValue->vt)'` and causes `SetValue()` to return `E_INVALIDARG`. Changing the variant type to `VT_UI4` fixes this issue.

Windows 10 retail build 19041.985

### DIFF
--- a/desktop-src/medfound/codecapi-avlowlatencymode.md
+++ b/desktop-src/medfound/codecapi-avlowlatencymode.md
@@ -12,7 +12,7 @@ Enables low-latency mode in a codec.
 
 ## Data type
 
-**ULONG** (**VT\_BOOL**)
+**USHORT** (**VT\_UI4**)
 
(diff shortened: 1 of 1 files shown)


PRMerger9:
@Xenonic : Thanks for your contribution! The author(s) have been notified to review your proposed change.

drewbatgit:
@Xenonic Thanks for catching this. I checked with the product team and it turns out that this is a bug with the H264 decoder. For other implementations, including the H264 encoder, the type is correctly implemented as VT_BOOL. I've updated the page with a warning to point out this inconsistency. So, I'm going to go ahead and close your PR. Thanks again for catching this bug.

PR #894, opened 2021-06-16
Update wer-settings.md

niderost:
I believe the LastFullLiveReport value is a FileTime value, not a SystemTime value, at least on my Windows 10, version 20H2, build 19042.1052 x64 edition.

### DIFF
--- a/desktop-src/wer/wer-settings.md
+++ b/desktop-src/wer/wer-settings.md
@@ -384,7 +384,7 @@ The maximum number of full live dumps that may be on disk at any given time. The
 
 **REG\_QWORD**
 
-A [SystemTime](/windows/desktop/api/minwinbase/ns-minwinbase-systemtime) indicating the last full live report time, for the system or a specific ReportType. This is used to calculate whether a policy threshold has been satisfied.
+A [FileTime](/windows/desktop/api/minwinbase/ns-minwinbase-filetime) indicating the last full live report time, for the system or a specific ReportType. This is used to calculate whether a policy threshold has been satisfied.
 
 </dd> <dt>
 


PRMerger16:
@niderost : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Karl-Bridge-Microsoft:
Hi @niderost - can you provide additional detail? I checked the Windows source code and the value appears to be a SystemTime.

Where did you get this info for your version of Windows that instigated this PR?

niderost:
> Hi @niderost - can you provide additional detail? I checked the Windows source code and the value appears to be a SystemTime.
> 
> Where did you get this info for your version of Windows that instigated this PR?

I accessed the registry value on my computer. The value is 0x1d7619d21a9dabc. When I try to decode this as a SystemTime, I get an invalid, nonsensical date. When I try to decode this as a FileTime, the value makes sense.

Maybe the SystemTime is converted into a FileTime during serialization when it is written to the Registry?

[FullLiveKernelReports.zip](https://github.com/MicrosoftDocs/win32/files/6668970/FullLiveKernelReports.zip)

0x1d7619d21a9dabc as SystemTime:

0x0000 0000 0000 0000 01d7 619d 21a9 dabc   // Assuming many leading 0 because size of _SYSTEMTIME struct is 8 WORDs (2-byte unsigned integer, see https://docs.microsoft.com/en-us/windows/win32/winprog/windows-data-types)

```
typedef struct _SYSTEMTIME {
  WORD wYear;          0x0000 =     0 -> doesn't make sense
  WORD wMonth;         0x0000 =     0 -> doesn't make sense
  WORD wDayOfWeek;     0x0000 =     0 -> doesn't make sense
  WORD wDay;           0x0000 =     0 -> doesn't make sense
  WORD wHour;          0x01d7 =   471 -> doesn't make sense
  WORD wMinute;        0x619d = 24989 -> doesn't make sense
  WORD wSecond;        0x21a9 =  8617 -> doesn't make sense
  WORD wMilliseconds;  0xdabc = 55996 -> doesn't make sense
} SYSTEMTIME, *PSYSTEMTIME, *LPSYSTEMTIME;

```

0x1d7619d21a9dabc as FileTime:

0x01d7 619d 21a9 dabc  // Assuming 1 leading 0 because size of _FILETIME struct is 2 DWORDs (4-byte unsigned integer)

```
typedef struct _FILETIME {
  DWORD dwLowDateTime;   21a9 dabc -> fits perfectly
  DWORD dwHighDateTime;  01d7 619d -> fits perfectly
} FILETIME, *PFILETIME, *LPFILETIME;
```
0x1d7619d21a9dabc = 132682041533258428 (x 100 ns intervals since Januari 1, 1601 (UTC)) = Tuesday, June 15, 2021 4:15:53am
(See https://www.silisoftware.com/tools/date.php?inputdate=132682041533258428+&inputformat=filetime)

Karl-Bridge-Microsoft:
Thanks for the detailed response. I think you might be correct with your serialization thought. I double checked the Windows source again and confirmed it's using SystemTime as stated in the topic. I'll close this PR, but please open another PR/Issue if you find more information on this functionality.

niderost:
I think you are misunderstanding my point on the serialization. This document is describing the content of the registry qword value "LastFullLiveReport", and that value IS a FileTime value, not a SystemTime struct. So the documentation should also mention FileTime, not SystemTime, since if you could put a SystemTime struct in the QWord (which is not possible since you would need 2 QWords for that), the subkey would not work as expected.

Now I assume the source code you are talking about is the source code of the Windows Error Reporting process. That source code might be using a SystemTime internally. However, it can't read/write the SystemTime directly from/into the registry since the registry doesn't support the SystemTime struct. Instead, some serialization/deserialization is needed. In the serialization / deserialization code (which you might have to search for outside of the Windows Error Reporting source code), you will find the conversion from SystemTime into a FileTime-alike struct or value that fits into the QWord in the registry. 

I.e., you should check the serialization / deserialization code details to find out how it is converting the SystemTime into a QWord. And also, you should check the registry key values on your own computer to validate that the value in the registry indeed is a FileTime and not a SystemTime. Finally, you should still accept the PR as that would make the documentation about what is in the registry subkey correct. The documentation now is incorrect as it's aim is not to describe what is in the WER source code (maybe a SystemTime, but I can't check) but what is in the registry subkey (definitely a FileTime, as you can check in the registry).

PR #906, opened 2021-06-24
Qualify outdated references to Freecell, Solitaire, Minesweeper

clandrew:
These games aren't included with clean installs of operating system as of Windows 8 and later. 

Games with similar functionalities are on the Microsoft Store, however this page isn't referring to those. To fix this it should qualify what OS they were part of.

This change pertains to the page [https://docs.microsoft.com/en-us/windows/win32/controls/cookbook-overview](https://docs.microsoft.com/en-us/windows/win32/controls/cookbook-overview)


PRMerger18:
@clandrew : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (approved):
Thank you!

PR #922, opened 2021-07-06
Update using-descriptors-directly-in-the-root-signature.md

AdamJMiles:
Clarified what types of buffers can and can't be bound directly in the root signature.


PRMerger20:

stevewhims (approved):

PR #928, opened 2021-07-09
Fixed a critical error in the documentation

Const-me:
According to the previous version of the documentation, column-major matrices (which happen to be the default in HLSL) store elements 14, 24, 34 twice, and don’t store elements 41, 42, 43 at all.

That’s not how they work.


PRMerger12:
@Const-me : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (approved):

Const-me:
I’m happy it’s merged and closed, but a “thank you” would be nice, still.
It was your fault after all, and an important one.

PR #935, opened 2021-07-16
Adding note about scope of queries supported

ChrisOfTheLewi:
A customer brought this up, and after conferring with our AD escalation engineers, this type of retrieval needs a base scope query (or a subtree search that returns one item).  Otherwise, results may, and after are, missing.  This is the long term behavior so I believe that documenting it rather than looking into changing it makes the the most sense.


PRMerger17:
@ChrisOfTheLewi : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #937, opened 2021-07-20
Update reserving-and-committing-memory.md

DBJDBJ:
Same as the whole WIN32 that is C, not C++ code. With proper UNICODE handling too. And using SEH on top of that. 

If the customer seeing this has never seen C she will be confused. 

In the year 2021, it is not rare to meet people using C++ who came from Python, JavaScript, C# etc...


PRMerger19:
@DBJDBJ : Thanks for your contribution! The author(s) have been notified to review your proposed change.

DBJDBJ:
To build this as C using CL it is very easy. For example: `cl /TC main.c`  
To build it as C++ requires more switches and then explanations on what and why are they.

GrantMeStrength:
Thanks for this PR!
To be consistent we try to stick to # and C++ in the docs.

DBJDBJ:
@GrantMeStrength I assume U meant to write  `C#` and `C++`?

PR #941, opened 2021-07-22
Fix reference to BC1 which should be BC4

fdwr:
The [Storing Compressed Data section](https://docs.microsoft.com/en-us/windows/win32/direct3d10/d3d10-graphics-programming-guide-resources-block-compression#storing-compressed-data) references BC1, saying it's an 8-byte block with 1-byte color values and 16 x 3-bit interpolation indices, but [BC1](https://docs.microsoft.com/en-us/windows/win32/direct3d10/d3d10-graphics-programming-guide-resources-block-compression#bc1) has 2-byte color values and 16 x 2-bit interpolation indices. It's actually [BC4](https://docs.microsoft.com/en-us/windows/win32/direct3d10/d3d10-graphics-programming-guide-resources-block-compression#bc4) which has 1-byte color values and 16 x 3-bit interpolation indices.

### DIFF
--- a/desktop-src/direct3d10/d3d10-graphics-programming-guide-resources-block-compression.md
+++ b/desktop-src/direct3d10/d3d10-graphics-programming-guide-resources-block-compression.md
@@ -46,9 +46,9 @@ The uncompressed data is laid out in memory sequentially and requires 16 bytes, 
 
 ### Storing Compressed Data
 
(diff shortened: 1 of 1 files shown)


PRMerger7:

stevewhims (approved):

PR #950, opened 2021-07-30
Fix misleading figure in descriptor tables doc

FreakingBarbarians:
I remember trying to create a texture in an upload heap and the debug layer yelling at me.
I've changed the relevant parts of the doc. I also provided a new diagram.
Not sure what the conventions are for storing image for diagrams so do let me know if it should be changed.
I've made an issue here https://github.com/MicrosoftDocs/feedback/issues/3669
also lmk if im just plain wrong I'll close both items :)


PRMerger15:
@FreakingBarbarians : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims:
Thanks for the PR and the GitHub issue. I'll run with the issue, do some investigation, and if we need to make a change to the topic then I'll make that in a separate PR. I understand the changes you're proposing in this PR (change "an upload" to "the default", in both the topic and the diagram).

So I'll go ahead and close this PR, and we'll pick up from here in the GitHub issue.

Thanks again!
Steve

PR #953, opened 2021-08-01
Change the message loop to exit on error and add comments

Zotyamester:
Even though GetMessage returns a BOOL type, it is actually possible that the function returns a negative value (due to the fact that BOOL is defined as an int in WinUser.h).
Thus, if we just hand the return value of GetMessage to the loop itself, it'll ignore the occasional errors.

Also, I grabbed the comment from desktop-src/LearnWin32/painting-the-window.md so it fills the empty break lines in the WindowProc function.


PRMerger9:
@Zotyamester : Thanks for your contribution! The author(s) have been notified to review your proposed change.

QuinnRadich (approved):

PR #955, opened 2021-08-03
Correct condition for scrollbar visibility

ivokabel:
The explanation of re-enabling the scrollbar by the system is in contradiction with the previous explanation of hiding/disabling of a scrollbar in the same paragraph. BOTH condition have to be fulfilled for a scrollbar to be enabled:

- Non-zero scroll area
- Page size that does not include the entire scroll range.


PRMerger9:
@ivokabel : Thanks for your contribution! The author(s) have been notified to review your proposed change.

jwmsft (approved):

PR #956, opened 2021-08-04
Fix run that doesn't end on a word boundary

MyNameIsTrez:
There was a discussion on StackOverflow about this run and how it is wrong, see it below for more detailed information on what this commit fixes.

The StackOverflow answers show proof that the padding byte is necessary, along with a link to nearly identical documentation that *does* include the padding byte.

I've also chosen to reword `each run must be aligned on a word boundary` to `each run must be zero-padded to end on a 16-bit word boundary`, as per these comments on that StackOverflow page:

> Presumably, this means that runs with an odd-numbered length must be zero-padded (although not specified, I assume words are 16 bits long in this context).

> Don't they mean that it has to start on a word-aligned address (in uncompressed form)? It's certainly vague, but that would be true in the example.

https://stackoverflow.com/questions/46349718/align-with-word-boundary-in-rle-bitmap-contradiction-in-microsoft-documentation

### DIFF
--- a/desktop-src/gdi/bitmap-compression.md
+++ b/desktop-src/gdi/bitmap-compression.md
@@ -26,13 +26,13 @@ When the **Compression** member of the bitmap information header structure is BI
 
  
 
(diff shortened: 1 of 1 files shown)


PRMerger17:
@MyNameIsTrez : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (approved):

PR #958, opened 2021-08-07
Fix description which states Delta bytes their vertical components go down, as they probably go up

MyNameIsTrez:
So I'm reading the [Microsoft Docs on bitmaps][1] and it has this part:

> Delta. The 2 bytes following the escape contain unsigned values indicating the horizontal and vertical offsets of the next pixel from the current position.

It doesn't specify whether horizontal means going left/right and it doesn't specify whether vertical means going up/down, but the `contain unsigned values` at least suggests that only one direction is possible for both of the axes.

There's an example at the bottom of that page that is supposed to clarify this ambiguity:

`[00 02 05 01]` means `move current position 5 right and 1 down`

So this sounds great and seems intuitive, but I think this is incorrect.

The [Wikipedia page on bitmaps][2] states this:

> Note that the bitmap data starts with the lower left hand corner of the image.

And [another source][3] I've found:

> The scan lines in the bitmap are stored from bottom up. This means that the first byte in the array represents the pixels in the lower-left corner of the bitmap and the last byte represents the pixels in the upper-right corner.

So you start reading from the bottom-left of a BMP, and you're then moving to the right, and once you reach the end of a row you move up.

This doesn't make sense with the `05 01 = 5 right and 1 down` from before though, because why would an (unsigned!) vertical value of 1 move you *down* when you start reading at the bottom row? I am not 100% sure that this is a mistake as I'm currently still trying to get my own RLE8 decompression code running, but this suggests that you're able to move *back*.

  [1]: https://docs.microsoft.com/en-us/windows/win32/gdi/bitmap-compression
  [2]: https://en.wikipedia.org/wiki/BMP_file_format#Compression
  [3]: http://www.martinreddy.net/gfx/2d/BMP.txt

### DIFF
--- a/desktop-src/gdi/bitmap-compression.md
+++ b/desktop-src/gdi/bitmap-compression.md
@@ -16,11 +16,11 @@ When the **Compression** member of the bitmap information header structure is BI
 
 
 
-| Value | Meaning                                                                                                                                                     |
-|-------|-------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| 0     | End of line.                                                                                                                                                |
-| 1     | End of bitmap.                                                                                                                                              |
-| 2     | Delta. The 2 bytes following the escape contain unsigned values indicating the horizontal and vertical offsets of the next pixel from the current position. |
+| Value | Meaning                                                                                                                                                |
+|-------|--------------------------------------------------------------------------------------------------------------------------------------------------------|
+| 0     | End of line.                                                                                                                                           |
+| 1     | End of bitmap.                                                                                                                                         |
+| 2     | Delta. The 2 bytes following the escape contain unsigned values indicating the offset to the right and up of the next pixel from the current position. |
 
 
 
--- a/desktop-src/gdi/bitmap-compression.md
+++ b/desktop-src/gdi/bitmap-compression.md
@@ -46,7 +46,7 @@ The bitmap expands as follows (two-digit values represent a color index for a si
 06 06 06 06 06 
 45 56 67 
 78 78 
-move current position 5 right and 1 down 
+move current position 5 right and 1 up 
 78 78 
 end of line 
 1E 1E 1E 1E 1E 1E 1E 1E 1E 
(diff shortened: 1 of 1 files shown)


PRMerger20:
@MyNameIsTrez : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims:
Thanks, I've reached out for a technical review of this change. Please stand by.

Many thanks!
Steve

MyNameIsTrez:
In case you could use an RLE8 BMP with a delta byte in it, [here you go](https://github.com/MyNameIsTrez/RLE8-BMP-Decompressor/blob/8344ab204c41766f839290e998ad1d403d2ac825/Input/rle8_with_delta.bmp).
You can verify it has a delta byte by looking for `00 02` with a hex editor.

MyNameIsTrez:
Here's visual proof of that BMP I linked that it should be `4 right, 1 up` and not `4 right, 1 down`.
example

MyNameIsTrez:
The reason there are also transparent spots here is because of "End of line." `00 00` bytes and "End of bitmap." `00 01` bytes causing early termination of a line/bitmap, so that's fine.

PRMerger14:
@MyNameIsTrez : Thanks for your contribution! The author(s) have been notified to review your proposed change.

MyNameIsTrez:
And in case there's still some doubt as to whether this goes for every single type of RLE8 BMP, I can say with 99.99% confidence that it does.

This is because I've finished writing my program that converts RLE8 BMPs to paletted PNGs, and I've used it on 430 RLE8 BMPs that were created by dozens of different people. I found that every single one of them adheres to the rule that the delta stands for (right, up), and ***never*** (right, down).

In case you need to check out the program you can find my repository for it [here](https://github.com/MyNameIsTrez/RLE8-BMP-Decompressor). Its `Tests` folder also contains a handful of RLE8 BMPs, which can be given to the program/a hex editor to verify my claims.

stevewhims:
Still awaiting a technical review of this change. Thanks for your patience.

-Steve

stevewhims:
We'll complete the PR now, thank you!

-Steve

PR #975, opened 2021-08-24
Fixed "control types" -> "control patterns"

wolfmanstout:
This used to say "UI Automation control types can be combined", which I think is incorrect because only one type can be assigned to an object (i.e. they cannot be combined). I think this ought to be "control patterns".


PRMerger7:
@wolfmanstout : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Karl-Bridge-Microsoft (approved):

PR #987, opened 2021-09-08
Update getqueuedcompletionstatusex-func.md

arobincaron:
This information was added to WaitForXXXObject but not here where it is also needed. This information is critical for developers to understand a breaking change made to Windows 8/2012.

There would be other API docs that might be in this information (GetQueuedCompletionStatus/GetOverlappedResultEx) but I cannot figure out how to find those in this API doc set.


PRMerger15:
@arobincaron : Thanks for your contribution! The author(s) have been notified to review your proposed change.

MatchaMatch:

PR #988, opened 2021-09-10
Updated MCN_GETDAYSTATE topic

Tanya-Solyanik:
NMDAYSTATE->prgDayState array is pre-allocated on the stack by the sender of this notification (calendar.cpp). Developer who handles this notification should not be allocating a new array that provides bitfields with bold dates, they should set bits in the given array.  See this topic for the correct handling of this notification: https://docs.microsoft.com/en-us/cpp/mfc/setting-the-day-state-of-a-month-calendar-control?view=msvc-160

### DIFF
--- a/desktop-src/Controls/mcn-getdaystate.md
+++ b/desktop-src/Controls/mcn-getdaystate.md
@@ -36,7 +36,7 @@ MCN_GETDAYSTATE
 *lParam* 
 </dt> <dd>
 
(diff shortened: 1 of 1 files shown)


PRMerger17:
@Tanya-Solyanik : Thanks for your contribution! The author(s) have been notified to review your proposed change.

jwmsft (approved):

PR #1000, opened 2021-09-15
Simplify WM_NCHITTEST table and sort items by value

DJm00n:


PRMerger3:
@DJm00n : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger13:
@DJm00n : Thanks for your contribution! The author(s) have been notified to review your proposed change.

DJm00n:
@Karl-Bridge-Microsoft any updates?

Karl-Bridge-Microsoft:
Hi @DJm00n - thanks for updating the topics. Re. the numeric ordering of the #Define values, even though the source code is in numeric order, from a readability/scanability perspective, I think it's better to leave them in alphabetical order. Most devs will use the #define value not the underlying numeric value. This will also introduce consistency issues between docs, for example: https://docs.microsoft.com/en-us/windows/win32/inputdev/wm-nchittest#return-value (although you could probably find counter examples too).

DJm00n:
@Karl-Bridge-Microsoft  This PR already contains changes to **WM_NCHITTEST** page - so both pages are containing the same table.

Also regarding readability I seeing alphabetical order as more confusing since related values like **HTLEFT**/**HTRIGHT**/**HTTOP**/**HTTOPLEFT**/**HTTOPRIGHT**/**HTBOTTOM**/**HTBOTTOMLEFT**/ **HTBOTTOMRIGHT** are completely mixed in table and its hard to do not miss any of them when I want to handle hittest to any window border.

I can remove actual int values from the table if you want but **WM_NCHITTEST** page already contains them.

Karl-Bridge-Microsoft:
@DJm00n - removing the int value won't affect alphabetical ordering and readability/scanability of the list. Thanks again for your contributions, but we've decided to leave the ordering as is.

PR #1016, opened 2021-09-27
UIA control pattern mappings should include Value for buttons

dlibby-:
On the Web, things like input type=file (which allows users to invoke a file picker) need to represent the chosen file via the value pattern. The user interface point is a button, so button should optionally support Value.


PRMerger12:
@dlibby- : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Karl-Bridge-Microsoft (approved):

PR #1017, opened 2021-09-28
Add a new architecture named LoongArch

kilaterlee:
Add LoongArch 32-bit and LoongArch 64-bit pe-coff support.

The application for IMAGE_FILE_MACHINE_LOONGARCH32 number is 0x6232, and
the application for IMAGE_FILE_MACHINE_LOONGARCH64 number is 0x6264.

Add two base relocation type for LoongArch, named
IMAGE_REL_BASED_LOONGARCH32_MARK_LA and IMAGE_REL_BASED_LOONGARCH64_MARK_LA.

Signed-off-by LiChao <lichao@loongson.cn>


PRMerger17:
@kilaterlee : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Karl-Bridge-Microsoft:
Hi @kilaterlee - I contacted the feature team re. this request and, based on their feedback, I have to reject this PR.

I suggest that if you'd like to submit this request through official product support channels, please try one of the following:

- [Feedback Hub](https://support.microsoft.com/en-us/windows/send-feedback-to-microsoft-with-the-feedback-hub-app-f59187f8-8739-22d6-ba93-f66612949332)
- [Microsoft Support](https://support.microsoft.com/contactus)

kilaterlee:
Dear @Karl-Bridge-Microsoft - Sorry, I think you may not understand what I mean. We are the owners of a new architecture called LoongArch, BIOS is UEFI must use the PE format, so I want to request the matchine number and describe the base relocation type for this new architecture in pe-format.md. Only when this PR is merged, UEFI will accept our patch. Can you help me?
In addition, I found that the RISC-V machine number and base address relocation type request has been accepted, can you tell me how they did it?
P.S. Our architecture Introduction URL: https://loongson.github.io/LoongArch-Documentation/README-EN.html, https://github.com/loongson/LoongArch-Documentation

Karl-Bridge-Microsoft:
Hi @kilaterlee - The feature team is looking into this. We'll update this thread when we have more info. Note - I'm still not sure how updating this doc will help you with your effort. This is just documentation about the product, it has not bearing on the product itself.

kilaterlee:
Hi @Karl-Bridge-Microsoft Thank you letting the feature team follow this case again. I think I know what you are worried about, but I have not other way to update yours product, so can you ask your feature team what we should do before updating this document? We found the RISC-V is in this document, what did they do before updating this document?
Thank you!

Karl-Bridge-Microsoft:
That question was also asked by the feature team when they reviewed the topic. I have no idea when that info was added and the available commit history for the topic doesn't help due to a doc platform migration a few years ago.

kilaterlee:
Dear @Karl-Bridge-Microsoft - Could you please help me see if the feature team responds to you?
By the way, we have only updated this document and will not affect Microsoft products, nor will we update Microsoft products. Can you ask if your feature team can be merged in this case?

spbrogan:
FYI - Progress is being made.

Whenthisworks1:
Will I never did so and would like to be removed...

kilaterlee:
Hi @spbrogan - I found that you are following the EDK2 project on GitHub, and I believe you know why I want to update this document. MS recommand that UEFI use the PE format, this is a historical issue. Therfore, if others architecture want to use UEFI, this document must be updated to add machine type and describe base relocation type for the new architecture. Please help me. Thank you very much!

VCZYK:
A few more questions on the new base relocation types:

- #define  IMAGE_REL_BASED_LOONGARCH32_MARK_LA    7
- #define  IMAGE_REL_BASED_LOONGARCH64_MARK_LA    8

If these two apply to LoongArch32 and LoongArch64 respectively, why can’t we assign same value to both base relocation types since there won’t be confusion?

As to the description on new base relocation types -- "This relocation is only meaningful when the machine type is LoongArch 32-bit (64-bit). The base address relocation applies a 32-bit (64-bit) absolute address through a combination of instructions",

- Does “combination of instructions” mean a “continuous sequence of instructions”?
- And how many bytes are there for such sequence of instructions?

kilaterlee:
Hi @VCZYK - For your questions:
_If these two apply to LoongArch32 and LoongArch64 respectively, why can’t we assign same value to both base relocation types since there won’t be confusion?_
* Use the same value maybe okay, but is it better to use one value? With two value, at least the 32-bit and 64-bit can be distinguished.

_Does “combination of instructions” mean a “continuous sequence of instructions”?_
* Yes, it means a continuous sequence of instructions.

_And how many bytes are there for such sequence of instructions?_
* In 32-bit, the sequence of instructions consists of 8 bytes, in 64-bit is 16 bytes. Our architecture is 4 bytes per instruction. In other words, in 32-bit is two instructions and 64-bit is four.

VCZYK:
Hi @kilaterlee 

Regarding using same value for the two base relocation types:

Because each of the two base relocation types only applies to one architecture only, i.e.,

- for LoongArch32 we would only see IMAGE_REL_BASED_LOONGARCH32_MARK_LA, and
- for LoongArch64 we would only see IMAGE_REL_BASED_LOONGARCH64_MARK_LA,

There won't be confusion if we assign both same value, like

- #define IMAGE_REL_BASED_LOONGARCH32_MARK_LA 7
- #define IMAGE_REL_BASED_LOONGARCH64_MARK_LA 7

Another question on the macro name, what does "MARK_LA" stand for?  Does it refer to the types of instructions that will need base relocation?

Existing base relocation macros are named as "IMAGE_REL_BASED_\<arch\>_\<type\>" and there is no underscore in "\<type\>".  So I'm curious what "MARK" and "LA" each refers to.

Thanks.

kilaterlee:
Hi @VCZYK - The issue of using the one value to refer to different architectures, we will discuss it internally.

_So I'm curious what "MARK" and "LA" each refers to._
* MARK_LA is defined by our ABI, refer to: https://loongson.github.io/LoongArch-Documentation/LoongArch-ELF-ABI-EN.html. There is a relocation type called R_LARCH_MARK_LA, which is used to mark the macro instruction "la", which means load address. If it is marked, "la" extends two or four consecutive instruction sequences, otherwise, it is uncertain to extend consecutive instruction sequences.

kilaterlee:
Dear @VCZYK and @spbrogan:
Is there has some news about this issue?

Thanks.

VCZYK:
Dear @kilaterlee ,

About using one value to refer to the two base relocation types for LoongArch (since there won't be confusion because one applies to LoongArch32 and the other applies to LoongArch64), you said you would discuss it internally.  Just wondering if you have reached an agreement and whether you are okay with using one value.

I think this is the only question we have before we can make progress on adding this into our PE spec.

Thanks.

kilaterlee:
Dear @VCZYK ,

Okay, if we have a result, I will tell you. Another question, if it is determined that one value can be used, it is just update the patch and have your team review it?

VCZYK:
Dear @kilaterlee,

We need to decide whether using one value is okay before we can move forward on modifying PE spec.  It is better if we can avoid publishing the PE spec first and then coming back to update it.  So please us know if using one value is sufficient or there is other technical reason that we have to use two values.

Thanks.

kilaterlee:
Dear @VCZYK,

As we have discussed, the result is that 32-bit and 64-bit use the same value, which is 8. Can I update this patch?

Thanks.

kilaterlee:
Dear @VCZYK,

I created a new PR, URL: https://github.com/MicrosoftDocs/win32/pull/1055
New PR changes:
1. Both 32-bit and 64-bit use the same value which is 8.
2. Modify the patch description.

Thanks.

VCZYK:
Dear @kilaterlee,

Thanks for the confirmation.  We will take care of updating the PE spec.

Thanks.

kilaterlee:
Dear @VCZYK ,

Okay, thank you for your attention to this issue. By the way, have you seen the new PR?

Thanks.

PR #1020, opened 2021-09-28
Delete private-communications-technology.md

jess-krynitsky:
This article is wildly obsolete, describes long unsupported code. Should be deleted.


PRMerger14:
@jess-krynitsky : Thanks for your contribution! The author(s) have been notified to review your proposed change.

hkhmsft:
@jess-krynitsky is there an existing article you think would be a good redirect for this?

jess-krynitsky:
I think you could redirect to "Protocols in TLS/SSL (Schannel SSP)"

PR #1021, opened 2021-09-30
Task 36552572: Edit signtool docs /fd /td info

jjhenry21:
Changes
1)Top Note: "A warning (error code 0, initially) will be thrown if /fd is not specified during signing and if /td is not specified during timestamping. In later versions of SignTool, the warning will become an error. " --> An error (error code 1) will be thrown if /fd is not specified during signing and if /td is not specified during timestamping.

2) All table rows where /td specified the note in the row needs to be updated to "An error is generated if /td is not provided while timestamping." (there is no default anymore <- does not need to go in doc"

3)All table rows where /fd is specified the note in the row needs to be updated to "An error  is generated if the/fd switch is not provided while signing."

### DIFF
--- a/desktop-src/SecCrypto/signtool.md
+++ b/desktop-src/SecCrypto/signtool.md
@@ -13,7 +13,7 @@ SignTool is a command-line tool that digitally signs files, verifies the signatu
 SignTool is available as part of the Windows SDK, which you can download from <https://developer.microsoft.com/windows/downloads/windows-10-sdk/>.
 
 > [!Note]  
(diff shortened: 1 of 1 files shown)


PRMerger-2:
@jjhenry21 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger12:
@jjhenry21 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

jjhenry21:
@lastnameholiu can you review and sing off on this PR? Thank you

hkhmsft:
Apologize for the delay. The changes should be live soon. Thanks!

PR #1022, opened 2021-09-30
Check the size of aac_profiles before accessing it in Tutorial: Encoding an MP4 File

austinorion:
This code sample incorrectly checks the size of h264_profiles instead of aac_profiles.


PRMerger19:
@austinorion : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #1025, opened 2021-10-04
Update obtaining-a-root-oid-from-an-iso-name-registration-authority.md

mvaneijken:
Both web.ansi.org and www.iso.ch are not valid anymore. Changed to ansi.org and www.iso.org.


PRMerger20:
@mvaneijken : Thanks for your contribution! The author(s) have been notified to review your proposed change.

MatchaMatch:

PR #1042, opened 2021-10-24
Bug 32265634: Misuse Potential in Signing API

jjhenry21:
added note: This function may return a count of duplicate signers and therefore may not be sufficient to avert attacks. We recommend using the sid (SignerIdentifier) field from SignerInfo to identify duplicate signers in a message.


PRMerger9:
@jjhenry21 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger7:
@lastnameholiu : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #1047, opened 2021-10-26
Clarify Unicode status of Windows Console

miniksa:
Modify bullet point 3 to clarify that the console does support Unicode, but some connected legacy tools might not.


PRMerger12:
@miniksa : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Karl-Bridge-Microsoft (approved):

PR #1056, opened 2021-11-01
small grammar fix

badasahog:


PRMerger-2:
@badasahog : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (commented):
In fact, I feel strongly that the current wording is correct here. The concept "graphics pipeline state" is being used here in an abstract sence as something that applies per user, per app, and even more finely-grained than that. So referring to "(most) graphics pipeline state" DOES feel appropriate. Using the definite articicle ("the") as in "(most of) the graphics pipeline state" feels appropriate only for referring to one particular instance of such state. Here we're referring to pipeline state in general.

An analogy would be "ocean currents". Talking about "(most of) the ocean currents" sounds to me like it'd be referring to (most of) the currents of one particular ocean (say, the Pacific, for example). And "(most) ocean currents" would refer to (most of) the currents of *all* of the World's oceans.

I'll go ahead and close the PR, since I think we're good with the topic as-is.

PR #1062, opened 2021-11-05
Fix time_t truncation in TimetToFileTime

CookiePLMonster:
Using `Int32x32To64`on time_t truncates the value. It can be observed via Compiler Explorer, where `movsxd  rax, DWORD PTR t$[rsp]` is used to load `time_t` to a register.
https://godbolt.org/z/T9h9vvec7

After my change, `imul    rax, QWORD PTR t$[rsp], 10000000` is used so the value is not truncated.

### DIFF
--- a/desktop-src/SysInfo/converting-a-time-t-value-to-a-file-time.md
+++ b/desktop-src/SysInfo/converting-a-time-t-value-to-a-file-time.md
@@ -1,33 +1,42 @@
 ---
-description: The time functions included in the C run-time use the time\_t type to represent the number of seconds elapsed since midnight, January 1, 1970. The following example converts a time\_t value to a file time.
+description: The time functions included in the C run-time use the **time_t** type to represent the number of seconds elapsed since midnight, January 1, 1970. The following example converts a **time_t** value to a [FILETIME](/windows/win32/api/minwinbase/ns-minwinbase-filetime).
 ms.assetid: f626c0b2-a5a1-475d-9a24-64e7b0407278
-title: Converting a time_t Value to a File Time
+title: Converting a time_t value to a FILETIME
 ms.topic: article
-ms.date: 05/31/2018
+ms.date: 12/03/2021
 ---
 
-# Converting a time\_t Value to a File Time
+# Converting a time\_t value to a FILETIME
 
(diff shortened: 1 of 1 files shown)


PRMerger7:
@CookiePLMonster : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims:
Thank you, I'm making enquiries with the feature team. Thanks for your patience!

-Steve

PRMerger3:

stevewhims (approved):

stevewhims:
Hi, @CookiePLMonster. Thank you for your change, and for explaining it. It was very helpful to see your before and after comparison in Compiler Explorer. All this kind of info is very much appreciated! :)

Thanks again!
Steve

desktop-src/SysInfo/converting-a-time-t-value-to-a-file-time.md
```diff
@@ -1,32 +1,42 @@
...
-    pft->dwLowDateTime = (DWORD) ll;
-    pft->dwHighDateTime = ll >>32;
+    ULARGE_INTEGER time_value;
+    time_value.QuadPart = (t * 10000000LL) + 116444736000000000LL;
+    pft->dwLowDateTime = time_value.LowPart;
+    pft->dwHighDateTime = time_value.HighPart;
 }
 ```
 
+After you've a [FILETIME](/windows/win32/api/minwinbase/ns-minwinbase-filetime), you can convert the value to system time using the [**FileTimeToSystemTime**](/windows/win32/api/timezoneapi/nf-timezoneapi-filetimetosystemtime) function.
```
CookiePLMonster, line 26: Missing "obtained" from the sentence.

CookiePLMonster (commented):
Thanks for the feedback @stevewhims! I re-reviewed the PR since I spotted a missing word in the changes you pushed prior to merging :)

desktop-src/SysInfo/converting-a-time-t-value-to-a-file-time.md
stevewhims: Good catch, @CookiePLMonster. I've fixed that in a new PR into the private repo. Ordinarily, that would be live within an hour or so, but we have an issue with the publishing pipeline. That should be cleared early next week, so both this PR and the one I just created should be live then.

Thanks!
Steve

PR #1066, opened 2021-11-06
Fix the over-translation of API names in Virtual-Key Codes

masaru-iritani:
# Summary

* Use inline code styles for API names.
* Split "Constant/Value" column into "Constant" column and "Value" column.
* Remove all HTML elements except `span` elements with `id` attribute.
* Remove unnecessary characters and spaces from tables.
* Remove `span` elements whose `id` attribute is `-`.

# Reason

* API names in Virtual-Key Codes page are over-translated in Japanese. For example, `VK_PAUSE` is translated to `VK _一時停止` in [Japanese](https://docs.microsoft.com/ja-jp/windows/win32/inputdev/virtual-key-codes#VK_PAUSE). [Microsoft Docs guideline](https://docs.microsoft.com/en-us/contribute/text-formatting-guidelines#code-style) encourages the use of code styles for code elements to leave them untranslated.
* Having two values in a cell makes the table complex and taller.
* The use of inline HTML is discouraged by [MD033](https://github.com/DavidAnson/markdownlint/blob/v0.24.0/doc/Rules.md#md033---inline-html). This change doesn't remove `span` elements with `id` attribute because some pages have been using those anchors.
* The current Markdown markup of tables is difficult to read because of a lot of redundant characters and spaces.
* Multiple elements set `-` as their ID. It's confusing and useless.

# Validation

* [x] [Docs Markdown](https://marketplace.visualstudio.com/items?itemName=docsmsft.docs-markdown) renders the updated page correctly.
* [x] [GitHub preview](https://github.com/masaru-iritani/win32/blob/21e55abd4d9c55b66d49b333e30dc33ac858542c/desktop-src/inputdev/virtual-key-codes.md) renders the updated content correctly.
* [x] [markdownlint](https://marketplace.visualstudio.com/items?itemName=DavidAnson.vscode-markdownlint) detects no new problem caused by this change.


PRMerger14:
@masaru-iritani : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Karl-Bridge-Microsoft (approved):

PR #1069, opened 2021-11-18
Update loopback-recording.md

rapsealk:
Hi. I was reading a document about [loopback-recording](https://docs.microsoft.com/en-us/windows/win32/coreaudio/loopback-recording) and found one malfunctioning command, which is
```C++
control mmsys.cpl,,1
```

When I tried that command, it does not work and throws only a `ParserError`. I guess it needs to be changed to the code below.
```ps1
control mmsys.cpl
```

Despite the command `control mmsys.cpl,1` works as well, but I think the following `,1` is not necessary.

Furthermore, I also change the *language type* of the script from `C++` to `ps1`.

Thank you for reading and please inform me if there is anything wrong.


PRMerger16:
@rapsealk : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #1106, opened 2022-01-19
max value of each 6 digit is 720885, not 720896.

Yamatohimemiya:
> Each group of 6 digits must be divisible by 11 and must be less than 720896.
This '720896' came from '65536 * 11'.

But 'unsigned short' range is 0-65535 and https://docs.microsoft.com/en-us/windows/win32/secprov/unlockwithnumericalpassword-win32-encryptablevolume also says ""less than"" 65536 (= 65535).

So correct max value is '65535 * 11' = 720885.


PRMerger15:
@HikaruDY : Thanks for your contribution! The author(s) have been notified to review your proposed change.

alvinashcraft (approved):
Thanks. I remember the update made to that other file. This looks good.

alvinashcraft:
#sign-off

PR #1111, opened 2022-01-25
Discussion: unicode API problem with relative path

umairsair:
I worked with CreateFileW API and long paths. When my path is like \\\\?\\C:\temp\longpath\longpath\\..\\..\test then it fails. If I remove ..\\.. then it works.

Is it documentation problem or I need to refer to some other documentation?

### DIFF
--- a/desktop-src/FileIO/naming-a-file.md
+++ b/desktop-src/FileIO/naming-a-file.md
@@ -142,7 +142,7 @@ The Win32 namespace prefixing and conventions are summarized in this section and
 
 For file I/O, the "\\\\?\\" prefix to a path string tells the Windows APIs to disable all string parsing and to send the string that follows it straight to the file system. For example, if the file system supports large paths and file names, you can exceed the **MAX\_PATH** limits that are otherwise enforced by the Windows APIs. For more information about the normal maximum path limitation, see the previous section [Maximum Path Length Limitation](#maximum-path-length-limitation).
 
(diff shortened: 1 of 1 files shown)


PRMerger12:
@umairsair : Thanks for your contribution! The author(s) have been notified to review your proposed change.

alvinashcraft:
Hi. Thanks for filing this issue. I will try the scenario and reply with my findings.

alvinashcraft:
Thanks for taking the time to let us know about this and we sincerely apologize for the delayed response. The scope of this feedback channel covers specific documentation fixes, but I can help redirect you to the right support channel to get an answer to your question. The community of experts and Microsoft Engineers over at [Microsoft Q&A](https://docs.microsoft.com/en-us/answers/topics/windows-api.html) offer timely answers to a multitude of technical questions surrounding Windows APIs.

A quick question though - Did you have a backslash between the second `longpath` and `..\..`? It seems like there should be another separator there.

If you do create a thread over at Microsoft Q&A, we would appreciate if you could share that link here so we can track the progress and incorporate any technical details into our documentation. Though I'm marking this Issue as closed now, please treat this thread as an open thread for follow-ups. If it's determined that there is a docs issue, we'll come back and make the appropriate changes. Thanks!

umairsair:
> A quick question though - Did you have a backslash between the second longpath and ..\..? It seems like there should be another separator there.

Its lost because of github comment formatting. I have updated first comment, backslash needed to be escaped.

PR #1112, opened 2022-01-25
Update wm-measureitem.md

Tanya-Solyanik:
`WParam` description in https://docs.microsoft.com/en-us/windows/win32/controls/wm-drawitem#parameters is accurate, it says ` If the message was sent by a menu, this parameter is zero.` WParam in `WM_MEASUREITEM` is used in the same way, however description says the reverse - `If the value is zero, the message was sent by a menu.` This is not correct. `WParam` value will also be set to zero if the combo or list box'control's resource ID is zero.

### DIFF
--- a/desktop-src/Controls/wm-measureitem.md
+++ b/desktop-src/Controls/wm-measureitem.md
@@ -39,7 +39,7 @@ WM_MEASUREITEM
 *wParam* 
 </dt> <dd>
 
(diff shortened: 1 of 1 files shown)


PRMerger3:
@Tanya-Solyanik : Thanks for your contribution! The author(s) have been notified to review your proposed change.

jwmsft (approved):

PR #1126, opened 2022-02-08
Improve description of descriptor table layout

bhouse-intel:
The current descriptor table layout description misleads the reader to think that the tables described must follow each other sequentially.  This is not the case.  The tables are start from a given offset within the heap but there ordering/positioning can vary.


PRMerger8:
@bhouse-intel : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (approved):
Thank you!

PR #1143, opened 2022-02-16
fix: typo on the flow chart

jukrb0x:
### Typo
`Input-assember Stage`
should be 
`Input-assembler Stage`

I would try to use some tool to change the typo since I don't have the original flowchart.


stevewhims:
Thanks, @jukrb0x, I've fixed the typo. Closing this suggestion.

PR #1144, opened 2022-02-18
right-left is misused vice versa

shahroozleon01:
"the left button is the primary button and the right is the secondary button." says in paragraph starting at line 19, but later says :"right button and left button to mean primary and secondary" which is opposite, assumed first one is the correct one.


PRMerger4:
@shahroozleon01 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

QuinnRadich (approved):

PR #1160, opened 2022-03-05
Slight correction to the value of XM_1DIV2PI

chrissinclair:
XM_1DIV2PI is defined in the header as: `XM_CONST float XM_1DIV2PI   = 0.159154943f;`, which looks more like 1/2pi rather than 2/pi, which would be 0.6366...


PRMerger16:
@chrissinclair : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #1164, opened 2022-03-07
Move GetProcessIdOfThread to correct section

henke37:
It operates on a thread handle so it belongs to the thread function section


PRMerger7:
@henke37 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Karl-Bridge-Microsoft (approved):

PR #1170, opened 2022-03-14
Clarify SEPARATOR section that the sample only works within MENU

fdwr:
Spent 20 minutes wondering why the .rc file wouldn't compile (as it didn't give an informative error message in the cmake build output), even though it looked exactly like the documentation. Thank goodness for other lost people on stackoverflow.

source doc: https://docs.microsoft.com/en-us/windows/win32/menurc/menuitem-statement


PRMerger12:

jwmsft (approved):

PR #1186, opened 2022-04-03
Fix definition of nit

hrxn:
A nit is 1 candela per square meter, not 1 candela per square centimeter..

https://en.wikipedia.org/wiki/Candela_per_square_metre


PRMerger8:
@Hrxn : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (approved):

PR #1187, opened 2022-04-06
Improve documentation of functions in Delay-Loaded DLLs section

jeremyd2019:
ResolveDelayLoadsFromDLL had incorrect information in the "Returns" section.

ResolveDelayLoadedAPI referred to the DelayLoadFailureHook function in the documentation for the FailureDllHook, when in fact DelayLoadFailureHook should be passed as the FailureSystemHook parameter.  (I'm still not sure that's clear).

All 3 functions were lacking documentation of the minimum Windows version required.

This whole section could probably use more work.  I noticed the "See also" section links all go to a download of retired technical documentation of Visual Studio 2003, when I think they were intended to refer to https://docs.microsoft.com/en-us/cpp/build/reference/linker-support-for-delay-loaded-dlls?view=msvc-170 (I would have updated them but I don't know the preferred way to make a permanent link, as opposed to whatever happened last time).

Even after all of this, the documentation is unclear how you're supposed to actually hook these APIs up.  The "Linker support" article goes into great detail on how it *used* to work, before these APIs were present.  Now that the OS provides these APIs it should be much simpler.

### DIFF
--- a/desktop-src/DevNotes/resolvedelayloadsfromdll.md
+++ b/desktop-src/DevNotes/resolvedelayloadsfromdll.md
@@ -63,7 +63,15 @@ Reserved; must be 0.
 
 ## Return value
 
-The address of the delay-load descriptor, if it is found; otherwise, **NULL**.
+Returns an **NTSTATUS** or error code.
+
+| Return code | Description |
+|--------------|--------------|
+| STATUS_SUCCESS | The operation completed successfully |
+| STATUS_DLL_NOT_FOUND | The Target DLL could not be found |
+
+The forms and significance of **NTSTATUS** error codes are listed in the Ntstatus.h header file available in the WDK, and are described in the WDK documentation.
+
 
 ## Requirements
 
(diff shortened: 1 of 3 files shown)


PRMerger14:
@jeremyd2019 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

desktop-src/DevNotes/resolvedelayloadsfromdll.md
```diff
@@ -63,19 +63,28 @@ Reserved; must be 0.
 
 ## Return value
 
-The address of the delay-load descriptor, if it is found; otherwise, **NULL**.
+Returns an **NTSTATUS** or error code.
+
+| Return code | Description |
+|--------------|--------------|
+| STATUS_SUCCESS | The operation completed successfully |
+| STATUS_DLL_NOT_FOUND | The Target DLL could not be found |
```
jeremyd2019, line 71: It didn't necessarily seem right to enumerate the NTSTATUS values, but Visual Studio's `dloadsup.h` does:
```c
    if (Status == STATUS_DLL_NOT_FOUND) {
        return HRESULT_FROM_WIN32(ERROR_MOD_NOT_FOUND);

    } else {
        return S_OK;
    }
```

That seems like a pretty good indication the only two choices for this function to return are DLL not found or success.

jeremyd2019: Well, I found another error code option... `STATUS_DELAY_LOAD_FAILED`.  It seems to return that if a delay imported procedure can't be found.  Actually, it returns that if the DLL can't be found too... I wonder what Visual Studio's code was thinking there...

jeremyd2019:
Would also be nice to document that the default failure hook raises exception 0x505 on failure (and what constant from the headers corresponds to that, presumably `ERROR_DELAY_LOAD_FAILED`).

QuinnRadich:
#sign-off

PR #1193, opened 2022-04-13
Fix some links in the Windows API index

guyharris:
Update some links not to point to previous-versions archived pages, and fix the capitalization of some other links (maybe the Web server that supplies them has a case-insensitive file system, but GitHub doesn't appear to, so they don't work when previewing.

This also removes an extra newline.

### DIFF
--- a/desktop-src/apiindex/windows-api-list.md
+++ b/desktop-src/apiindex/windows-api-list.md
@@ -55,7 +55,7 @@ The Windows UI API create and use windows to display output, prompt for user inp
 ## Windows environment (Shell)
 
 -   [Windows Property System](../properties/property-system-reference.md)
--   [Windows Shell](/previous-versions/windows/desktop/legacy/ff521731(v=vs.85))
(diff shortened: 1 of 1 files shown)


PRMerger8:
@guyharris : Thanks for your contribution! The author(s) have been notified to review your proposed change.

guyharris:
While you're at it, since I don't see an Issues list, both the Windows API index page, and the [top-level Desktop app technologies page](https://github.com/MicrosoftDocs/win32/blob/docs/desktop-src/desktop-app-technologies.md) should perhaps sync up their lists of top-level topics and subtopics, and then remove more previous-versions links, in cases where there is currently-maintained documentation, and perhaps mark cases where is no currently-maintained documentation as older APIs.

(This all came up when I was updating [the Wikipedia page for the Windows API](https://en.wikipedia.org/wiki/Windows_API), the "Overview" section of which needs a major update.)

GrantMeStrength:
Thank you!

PR #1207, opened 2022-04-26
Document supported bitrates for 5.1 channel audio

auorion:
Experimentation, along with examining the output GetOutputAvailableType shows that the only supported types that contain 6 audio channels must have an MF_MT_AUDIO_AVG_BYTES_PER_SECOND that is either 72000, 96000, 120000, or 144000. These values are 6x the supported values for mono and stereo.

This is confirmed through code inspection and consultation with the API owner.


PRMerger16:
@auorion : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #1220, opened 2022-05-04
Update dynamic-link-library-best-practices.md

jasonepperly:
Removing the "reason" after the rule that says don't call registry functions regarding advapi32. We are getting a lot of developers that read the reason and try to mitigate the reason through various means. This reason is only one of many reasons you cannot use registry functions in dllmain, e.g., if you use a pre-defined handle the registry code will use a critical section which can lead to deadlocks.


PRMerger-2:
@jasonepperly : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (approved):

PR #1230, opened 2022-05-12
Update using-sspi-with-a-windows-sockets-client.md

swatish17:
Running this sample code throws SEC_E_INVALID_HANDLE (0x80090301) error. 

The reason seems to be the improper usage of _SecHandle variables. The global variable hcText is initialized in DoAuthenticate() method, whereas the local variable hCtxt is used in the main() method (which is not initialized). 

Fix : Replaced the local variable hCtxt with the global variable hcText. Also some changes with respect to SspiExample.h. (typecast to GenClientContext parameter to SEC_WCHAR* and MyHandleError parameter to const char*)

TODO: Also please remove the unused _SecHandle parameter from the ConnectAuthSocket() in this file as well as from SspiExample.h.

### DIFF
--- a/desktop-src/SecAuthN/using-sspi-with-a-windows-sockets-client.md
+++ b/desktop-src/SecAuthN/using-sspi-with-a-windows-sockets-client.md
@@ -40,6 +40,8 @@ This sample program uses limited error handling.
 #include <winsock.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <schannel.h>
+#include <Security.h>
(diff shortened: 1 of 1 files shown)


PRMerger18:
@swatish17 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

alvinashcraft (approved):
The changes look good here, and I will create an internal work item to address the unused parameter in the header file and client code.

alvinashcraft:
#sign-off

PR #1232, opened 2022-05-15
Fix wParam usage in the example for WM_SYSCOMMAND

CookiePLMonster:
In `WM_SYSCOMMAND` the Remarks section states:
> In WM_SYSCOMMAND messages, the four low-order bits of the wParam parameter are used internally by the system. To obtain the correct result when testing the value of wParam, an application must combine the value 0xFFF0 with the wParam value by using the bitwise AND operator.

This PR updates the example code that was **not** doing this on an assumption that it's this remark that is correct and not the example code. Should this be the other way around, I think that remark should be removed instead, but looking at the possible `wParam` parameters, there should be no harm in suggesting doing so even if it's not needed (anymore).

### DIFF
--- a/desktop-src/menurc/wm-syscommand.md
+++ b/desktop-src/menurc/wm-syscommand.md
@@ -30,7 +30,7 @@ A window receives this message when the user chooses a command from the **Window
 
 ```cpp
  case WM_SYSCOMMAND:
-        if (wParam == SC_CLOSE)
+        if ((wParam & 0xFFF0) == SC_CLOSE)
         {
             EndDialog (hDlg, TRUE);
             return(TRUE);


PRMerger16:
@CookiePLMonster : Thanks for your contribution! The author(s) have been notified to review your proposed change.

CookiePLMonster:
It's been nearly a year - is anything wrong with this suggestion?

riverar (approved):

CookiePLMonster:
Another year, another bump - is this the case of the docs being wrong while the example code is correct, or vice versa?

riverar:
@stevewhims @alvinashcraft Low hanging fruit here, mind signing off here?

stevewhims:
> @stevewhims @alvinashcraft Low hanging fruit here, mind signing off here?

I'd say the next step is to contact the product group, and confirm whether the remark or the code snippet is correct (the writer-owner will have contact info).

riverar:
I'm external. Do you have a flowchart or compass on how to resolve Win32 docset issues like this? I can start looping in folks here but that'd be noisy.

jwmsft (approved):

jwmsft:
Sorry for the delay on this, it can be difficult to find people to confirm updates to old code examples like this. (This is from a [Windows Classic Sample](https://github.com/microsoft/Windows-classic-samples/blob/1d363ff4bd17d8e20415b92e2ee989d615cc0d91/Samples/Win7Samples/winbase/registry/RegExplorer.c) that was last touched 8 years ago and also needs an update.) However, after searching internal codebases for examples, I'm confident that the updated example is the correct usage. Thanks for submitting this.

PR #1233, opened 2022-05-16
Changed I switch to lowercase i

martingbrown:
When rendered in Edge, the /I (uppercase i) switch looks like it is a lower case L which is confusing. All the other switches are lower case, the example below the table uses a lower case i and the /? text produced by the tool uses a lower case i, as such this change makes things consistent. It might be an idea if docs.microsoft.com switched to a font where the lowercase l and uppercase I didn't look identical, but I suspect that is a bigger change.


PRMerger15:
@martingbrown : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (approved):

PR #1269, opened 2022-06-16
Compile error

shd8698766:
The error is: Error (active)	E0167	argument of type "const wchar_t *" is incompatible with parameter of type "LPWSTR"	
It should be a LPCWSTR and not LPWSTR

Same thing here: https://docs.microsoft.com/en-us/windows/win32/dlls/creating-a-simple-dynamic-link-library?ns-enrollment-type=Collection&ns-enrollment-id=bookmarks   where the DLL "MyPuts" was created.


PRMerger8:
@shd8698766 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (approved):

PR #1270, opened 2022-06-16
It's LPCWSTR and not LPWSTR

shd8698766:
So it can work with its example: https://docs.microsoft.com/en-us/windows/win32/dlls/using-run-time-dynamic-linking


PRMerger5:
@shd8698766 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (approved):

PR #1280, opened 2022-06-27
Update reporting-an-event.md

ocpalo:
Line 115: new line needed for mc compiler

Line 288-293: rhs is a const w_char* but lhs was LPWSTR which is w_char*. It needs to be LPCWSTR. Otherwise code could not be compiled. Hence pInsertString have to be LPCWSTR array. 

Line 312: ReportEvent last argument needs LPVOID but LPWCSTR is not suitable. Const_cast needed.

### DIFF
--- a/desktop-src/EventLog/reporting-an-event.md
+++ b/desktop-src/EventLog/reporting-an-event.md
@@ -113,6 +113,7 @@ SymbolicName=GALLONS_UNITS
 Language=English
 gallons%0
 .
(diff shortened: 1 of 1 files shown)


PRMerger15:
@ocpalo : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Karl-Bridge-Microsoft (approved):

PR #1282, opened 2022-06-30
Fixed the table formatting and contained code

ThubLives:
This table was a bit of a mess; having vertical pipes in the code examples was not handled well by the table markup, which uses vertical pipes.  The HTML tags intended to preserve the example code didn't have the desired effect.

I replaced the code and pre tags with normal markdown inline code markup.  I also escaped all pipe characters in the table whether or not they are code examples since the table formatting was taking precedence over inline code preservation.

I'm not convinced that putting code examples with long lines inside of a table is the best way to present this information due to how hard it is to read very wide tables, but at least now all the information is actually being displayed now.  Maybe the information would be easier to understand without the table and instead described under subheadings so the examples can take the whole width of the viewport if needed.


PRMerger5:
@ThubLives : Thanks for your contribution! The author(s) have been notified to review your proposed change.

drewbatgit:
#sign-off Thanks for your contribution!

PR #1283, opened 2022-07-01
EndQuery is supported by compute command lists

mitko0003:
According to the query type support table in https://microsoft.github.io/DirectX-Specs/d3d/CountersAndQueries.html#queries, EndQuery method should be supported by compute command lists.


PRMerger17:
@mitko0003 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (approved):

PR #1297, opened 2022-07-24
Fix EN_CHANGE redirection

IvanVenkov712:
EN_CHANGE redirects to windowless Rich Edit control EN_CHANGE but it should redirect to Edit control EN_CHANGE


PRMerger17:
@IvanVenkov712 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #1298, opened 2022-07-26
ETW Autologgers are not configured with EnabeTraceEx2

jdu2600:
Hi.

I suggest that this paragraph should be deleted. It appears to have been copied verbatim from [configuring-and-starting-an-event-tracing-session](https://docs.microsoft.com/en-us/windows/win32/etw/configuring-and-starting-an-event-tracing-session) and does not seem to be relevant to Autologger configuration.

Specifically, this paragraph references `EnableTraceEx2` and the previous paragraphs explicitly states that Autologger sessions are configured by `EnableTraceEx` plus the table only references parameters relevant to `EnableTraceEx`.

That said, it would be awesome if ETW Autologger sessions could be configured via `EnableTraceEx2`. If that is the case then can you please update the table with the necessary registry keys to set. :-)

### DIFF
--- a/desktop-src/ETW/configuring-and-starting-an-autologger-session.md
+++ b/desktop-src/ETW/configuring-and-starting-an-autologger-session.md
@@ -117,8 +117,6 @@ The AutoLogger sessions increase the system boot time and should be used sparing
 
 To stop an AutoLogger session, call the [**ControlTrace**](/windows/win32/api/evntrace/nf-evntrace-controltracea) function. The session name that you pass to the function is the name of the registry key that you used to define the session in the registry.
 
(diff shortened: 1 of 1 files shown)


PRMerger12:
@jdu2600 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

jdu2600:
Background on a use case for Autologger event id filter support is here - https://github.com/jdu2600/RpcRegistrationMonitor/blob/c48371e5f07d4a20474ee2fb6d1ccc48b57fb59b/Program.cs#L8-L12

cc: @Philts

Karl-Bridge-Microsoft (approved):

PR #1301, opened 2022-08-05
Update complete-server-code.md  --- fixed comment

pea-sys:
Create a SOCKET for connecting to server → Create a SOCKET for the server to listen for client connections

### DIFF
--- a/desktop-src/WinSock/complete-server-code.md
+++ b/desktop-src/WinSock/complete-server-code.md
@@ -67,7 +67,7 @@ int __cdecl main(void)
         return 1;
     }
 
-    // Create a SOCKET for connecting to server
+    // Create a SOCKET for the server to listen for client connections
     ListenSocket = socket(result->ai_family, result->ai_socktype, result->ai_protocol);
     if (ListenSocket == INVALID_SOCKET) {
         printf("socket failed with error: %ld\n", WSAGetLastError());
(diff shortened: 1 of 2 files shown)


PRMerger19:
@pea-sys : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims:

stevewhims (commented):
For the wording in the first file, I'll go ahead and make that change in another PR, thank you! :)

For the changes in the second file, there are so many occurrences of things like escaped underscores, and HTML tables, that it doesn't scale if we have customers fix them manually one by one, and then have writers double-check every change on by one. Across the docs it's likely there'd be hundreds of thousands, or millions, of such edits. Lining up markdown tables like that isn't necessary, either (I generally remove all doubled spaces and all doubled hyphens). And we don't want to perpetuate the use of HTML in markdown files by adding CSS styling; we should where possible be replacing HTML with markdown. I think that all changes of this sort are best done in an automated fashion by the writing team.

stevewhims:
I've taken the change to the first file and I've used it in a separate PR. That change is live now. But I'll go ahead and close this PR without taking the changes to the second file (for the reasons described above).

Thanks!
-Steve

PRMerger-2:
@pea-sys : Thanks for your contribution! The author(s) have been notified to review your proposed change.

pea-sys:
＠[stevewhims](https://github.com/stevewhims)
Thank you for your review

PR #1303, opened 2022-08-05
Fixed various

jasonepperly:
Some parts of the article referred to lpReserved, while others used lpvReserved. Need to be consistent for clarity. Also updated sample code to make sure the process detach checked the lpvReserved and if it was not NULL to avoid all cleanup  code (as detailed in the remarks of the artlcle)


PRMerger18:
@jasonepperly : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (approved):
Thank you, consistency is good so this is better. Strictly speaking, we shouldn't be using Hungarian notation, so perhaps the variable should be named *reserved*. But then for consistency we'd need to change all the other variables in this example and in many thousands of others. :)

PR #1310, opened 2022-08-16
Add tokenGroups attribute to user security properties

simon-biber:
The tokenGroups attribute is a useful attribute for obtaining a list of group memberships in just two LDAP queries: the first to get the list of group SIDs from the tokenGroups attribute of the user, the second using that list of SIDs to get the name attribute of each group. It avoids the need to make multiple searches to expand the primaryGroupId attribute and recursively expand the memberOf attribute.

### DIFF
--- a/desktop-src/AD/security-properties.md
+++ b/desktop-src/AD/security-properties.md
@@ -235,11 +235,11 @@ The [**scriptPath**](/windows/desktop/ADSchema/a-scriptpath) attribute specifies
 <span id="tokenGroups"></span><span id="tokengroups"></span><span id="TOKENGROUPS"></span>[**tokenGroups**](/windows/win32/adschema/a-tokengroups)
 </dt> <dd>
 
-The [**tokenGroups**](/windows/desktop/ADSchema/a-tokengroups) attribute is a multi-valued attribute that contains the SID of all groups of which the user is a direct and indirect member, including for the primary group. This attribute can only be retrieved if a Global Catalog (GC) server is present to retrieve the transitive reverse memberships.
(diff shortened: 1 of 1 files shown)


desktop-src/AD/security-properties.md
```diff
@@ -232,6 +232,17 @@ The [**scriptPath**](/windows/desktop/ADSchema/a-scriptpath) attribute specifies
 
 </dd> <dt>
 
+<span id="tokenGroups"></span><span id="tokengroups"></span><span id="TOKENGROUPS"></span>[**tokenGroups**](/windows/win32/adschema/a-tokengroups)
```
alvinashcraft, line 235: This looks good, but I have a minor suggestion. The "/windows/desktop" path now redirects to "/windows/win32". For any new or updated content, we should use the new path to avoid redirects. For tokenGroups, it would be /windows/win32/adschema/a-tokengroups
Thanks!

alvinashcraft (approved):
Thanks for your contribution and for updating the win32 path. Looks good!

simon-biber:
This seems to have broken the rendering of the next item at https://docs.microsoft.com/en-us/windows/win32/ad/security-properties 

image

alvinashcraft:
I see that now. I don't see any problems in the markdown, but pages with HTML mixed in can have unpredictable output at times. This entire page could use a refresh with the spans/links replaced by headers.

```
## unicodePwd

The [**unicodePwd**](/windows/desktop/ADSchema/a-unicodepwd) attribute is the user password.
```

I'll see if I can get some time this week to update the whole page to remove the span, dt, and dd tags.

alvinashcraft:
Thanks for the heads-up about this issue @simon-biber. I've completed another PR to fix the issue and clean up the formatting. The updates are live now.

PR #1312, opened 2022-08-16
Update sm5-object-rwstructuredbuffer-decrementcounter.md

m1yur1:
Fix title and api_location.


stevewhims:
Thanks for the PR, @m1yur1!

I'm reluctant to take changes to metadata via PRs. Metadata is very sensitive. It's purpose is *primarily* (although not exclusively) for internal tools to use, and not for readers to see. And it's very easy to break metadata by changing it.

The changes here don't seem to be valuable enough to outweigh the inherent risk. Certainly I'd prefer not to change api-location unless I'm doing it myself. I can change the description at some point if I get bandwidth.

Hope that's all ok. :)

Thanks again!
-Steve

PR #1314, opened 2022-08-20
Use codeblock to show signtool command

norihiro:
This PR fixes command line example to use codeblock instead of bold text.

Actual issue is that the [Japanese translation](https://docs.microsoft.com/ja-jp/windows/win32/seccrypto/adding-time-stamps-to-previously-signed-files) of this document has a wrong command line as below. A space between the timestamp server and the file name is removed.
```
signtool timestamp -t https://timestamp.digicert.comMyControl.exe
```


alvinashcraft (approved):
Looks good. Thanks for fixing this formatting issue!

PR #1318, opened 2022-08-26
Update wm-displaychange.md

scottsmith-im:
I'm creating a Pull Request for this after submitting. I'm new at this, but I suspect that without submitting the Pull Request, I'm just committing the change to my local branch, and no one at Microsoft is going to see it. If I'm wrong about that, my apologies; please let me know.

Cheers

---

This documentation needs to be updated to include multiple-display setups. For a single-display system, the usage is clear. For multi-display systems, there's no obvious way to relate this message's **width**, **height** and **bpp** with a _specific_ display on the system. Since multiple displays can have identical modes set, it's not clear how to use this information effectively.

### DIFF
--- a/desktop-src/gdi/wm-displaychange.md
+++ b/desktop-src/gdi/wm-displaychange.md
@@ -48,6 +48,18 @@ The high-order word specifies the vertical resolution of the screen.
 
 This message is only sent to top-level windows. For all other windows it is posted.
 
+**BEGIN**
+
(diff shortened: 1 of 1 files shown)


stevewhims:
Thanks, @scott-smith-immersedvr!

FYI, I can't take this PR, because doing so would embed customer feedback into the documentation. :)

GitHub issues are the way to give feedback about incorrect or missing info in the documentation. It would be great if you could visit https://github.com/MicrosoftDocs/win32-pr/issues, and create a GitHub issue with this content, plus a link to the topic in question (which in this case is https://docs.microsoft.com/windows/win32/gdi/wm-displaychange).

I'll go ahead and close the PR.

Thanks!
-Steve

PR #1333, opened 2022-09-07
Remove content about Shell Extensions Approved registry key

ChrisGuzak:
this functionality was removed prior to Windows 7.


luedersj:
> this functionality was removed prior to Windows 7.

Instead of just deleting should we note that this is no longer required as of Win7?

RobertStPierre (approved):
Thanks for updating this Chris!

ChrisGuzak:
> > this functionality was removed prior to Windows 7.
> 
> Instead of just deleting should we note that this is no longer required as of Win7?

I don't think so. it really never worked so no reason to document it at all.

ChrisGuzak:
@jwmsft can you complete this?

PR #1335, opened 2022-09-08
Update about-keyboard-input.md

DJm00n:
Add virtual key code list link


DJm00n:
@Karl-Bridge-Microsoft looks like Scan Code Table on this page is missing "Note 1" etc descriptions after the table. Is it supposed to be like this?

Karl-Bridge-Microsoft (approved):

PR #1345, opened 2022-09-22
Title Case Composition Swapchain

BL1NDX3N0N:
Casing for composition swapchain has been updated to adhere with the rest of the document.


stevewhims:
Thanks, @BL1NDX3N0N. When used in the middle of a sentence (so that casing issues are eliminated), it's "composition swapchain". So it looks like this content is already correct, so I'll close the PR. Oh, and our guidelines call for sentence case in titles (although there are some places where that's not adhered to and needs to be fixed). :)

Thanks!
-Steve

BL1NDX3N0N:
I will be sure to keep an eye out for that. Have a good day.

PR #1351, opened 2022-09-26
Fixed broken link

jasonepperly:
Fixing broken link in the summary section.


stevewhims:
The original url was correct, but it was easy to miss the fact that you then had to search for the whitepaper at the destination (so I fixed that here). The PR was changing the url to a destination where, as far as I could see, the whitepaper wasn't available (so I changed the url back).

-Steve

PR #1353, opened 2022-09-28
Changed parameter name for callbacks

DavidWindham:
Was named pNotify, but RegisterEndpointNotificationCallback and UnregisterEndpointNotificationCallback have the parameter of "pClient"


drewbatgit:
@DavidWindham - Thanks for catching this!

PR #1354, opened 2022-09-29
Please update PlayMidiFile example for x64

petr-maly:
Current example causes access violation on x64 platform, because the pointers are truncated. Please, update example for 64bit. Thanks and regards,
Petr Maly


drewbatgit:
#sign-off

drewbatgit (approved):

drewbatgit:
#sign-off

PR #1361, opened 2022-10-07
Update management-filtering-layer-identifiers-.md

shankarseal:
The text for connect/bind redirect layers is corrected.


stevewhims (approved):

junwumsft (commented):
L626, the old text indicates there is a difference between TCP and UDP. Can UDP modify the remote address and port?

shankarseal:
> L626, the old text indicates there is a difference between TCP and UDP. Can UDP modify the remote address and port?

Yes, it can.

PR #1365, opened 2022-10-08
Update manifest-file-schema.md

dragon-dreamer:
Move some elements to correct namespaces. Tested namespaces on a real exe file. If those elements are put to other namespaces, exe file does not load.

Some elements are also documented with the corresponding correct namespaces:
- `highResolutionScrollingAware`, `ultraHighResolutionScrollingAware` should be in `http://schemas.microsoft.com/SMI/2013/WindowsSettings`: [link](https://learn.microsoft.com/en-us/windows/compatibility/precision-touchpad-devices#solution).


drewbatgit:
Thanks for catching this and making the fix!

PR #1372, opened 2022-10-13
updating document details

ryanriesMSFT:
updating document details that dirsync cookies are not always guaranteed to be internally consistent among different versions of Windows, i.e., you may not be able to reuse a cookie you got from a Server 2012R2 LDAP server, and reuse that cookie on a Server 2022 server.


desktop-src/AD/polling-for-changes-using-the-dirsync-control.md
```diff
@@ -37,6 +37,8 @@ Also, be aware that:
...
+    
+-   It is also possible the other DC rejects the cookie returned from the original DC. The search generates an LDAP error on the server like "0000203D: LdapErr: DSID-xxxxxxxx, comment: Error processing control, data 0", and the client application may generate an error such as "System.DirectoryServices.Protocols.DirectoryOperationException: A protocol error occurred." This may for example happen when the cookie is older and the internal contents of the cookie are expected to be different when processed by an LDAP server running a different version of Windows. The cookie is an opaque structure and is not guaranteed to be structurally consistent among all Windows OS versions. The client application should handle this case and retry with a full sync if this error is encountered.
```
alvinashcraft:
```suggestion
-   It is also possible the other DC rejects the cookie returned from the original DC. The search generates an LDAP error on the server like "0000203D: LdapErr: DSID-xxxxxxxx, comment: Error processing control, data 0", and the client application may generate an error such as "System.DirectoryServices.Protocols.DirectoryOperationException: A protocol error occurred." This may happen, for example, when the cookie is older, and the internal contents of the cookie are expected to be different when processed by an LDAP server running a different version of Windows. The cookie is an opaque structure and is not guaranteed to be structurally consistent among all Windows OS versions. The client application should handle this case and retry with a full sync if this error is encountered.
```

Thanks for submitting this. Just a minor suggested tweak to the wording.

ryanriesMSFT: OK cool, looks good, thanks

alvinashcraft:
```suggestion
-   It is also possible the other DC rejects the cookie returned from the original DC. The search generates an LDAP error on the server like "0000203D: LdapErr: DSID-xxxxxxxx, comment: Error processing control, data 0", and the client application may generate an error such as "System.DirectoryServices.Protocols.DirectoryOperationException: A protocol error occurred." This may happen, for example, when the cookie is older, and the internal contents of the cookie are expected to be different when processed by an LDAP server running a different version of Windows. The cookie is an opaque structure and is not guaranteed to be structurally consistent among all Windows OS versions. The client application should handle this case and retry with a full sync if this error is encountered.
```

Sorry, it looks like the change didn't apply yesterday. I'm going to commit the change now and then approve the PR.

alvinashcraft (approved):
Looks good. Thanks again!

PR #1374, opened 2022-10-14
Remove erroneous statement from a past age

SphinxKnight:
It has been a while since `try...catch` has been implemented in JavaScript/ECMAScript (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch#specifications )


stevewhims:
I made a bit of a tweak to remove *only* the incorrect info. :)

stevewhims (approved):

PR #1384, opened 2022-10-18
Update certificate chain ordering for schannel verification

nmoinvaz:
See https://github.com/curl/curl/issues/9706. Have not gotten confirmation that this was an intended change by Microsoft. Some updates for Windows 11 have been showing there have been issues in the SSL stack.


alvinashcraft:
Hi @nmoinvaz. Thanks for the PR/question. I've emailed someone in engineering to find out if this change in behavior was intended and what we should document here. I'll follow up when I get more information.  - Alvin

nmoinvaz:
@alvinashcraft any updates?

alvinashcraft:
@nmoinvaz Here are the responses I got:

"The text is correct: the leaf certificate is the last in the certificate chain. Certificate chain order is determined by the order in which certs sign each other, not by the order certificates are listed in a particular SSPI attribute’s memory buffer."

and

"Perhaps a useful sentence to add is something like: 'Certificate ordering in the SSPI context buffer does not imply any certificate chaining relationship'."

desktop-src/SecAuthN/manually-validating-schannel-credentials.md
```diff
@@ -10,7 +10,7 @@ ms.date: 05/31/2018
...
+To manually validate the server certificate, you must first get it. Use the [**QueryContextAttributes (General)**](/windows/win32/api/sspi/nf-sspi-querycontextattributesa) function and specify the SECPKG\_ATTR\_REMOTE\_CERT\_CONTEXT attribute value. This attribute returns a [**CERT\_CONTEXT**](/windows/win32/api/wincrypt/ns-wincrypt-cert_context) structure containing the certificate supplied by the server. This certificate is called the leaf certificate because it is the last certificate in the certificate chain and is farthest away from the [*root certificate*](../secgloss/r-gly.md). Starting in Windows 11 22H2 22621, the leaf certificate appears first in the chain?
```
alvinashcraft: Suggested change based on response from engineering team:

```suggestion
To manually validate the server certificate, you must first get it. Use the [**QueryContextAttributes (General)**](/windows/win32/api/sspi/nf-sspi-querycontextattributesa) function and specify the SECPKG\_ATTR\_REMOTE\_CERT\_CONTEXT attribute value. This attribute returns a [**CERT\_CONTEXT**](/windows/win32/api/wincrypt/ns-wincrypt-cert_context) structure containing the certificate supplied by the server. This certificate is called the leaf certificate because it is the last certificate in the certificate chain and is farthest away from the [*root certificate*](../secgloss/r-gly.md). However, certificate ordering in the SSPI context buffer does not imply any certificate chaining relationship.
```

alvinashcraft (approved):
This wording looks good. Thanks for taking the time to improve the docs!

nmoinvaz:
Thanks, I have made some changes and added what you proposed. I think the way it was previously written did imply that the first certificate would be the leaf certificate.

nmoinvaz:
I think what is still missing is instructions on how to get the leaf certificate from the store. I'm not sure what is the best way to enumerate the certificate store in a particular order to find leaf or root.

alvinashcraft:
Is that covered in the [Listing the Certificates in a Store](https://learn.microsoft.com/en-us/windows/win32/seccrypto/example-c-program-listing-the-certificates-in-a-store) topic?

"The following example code lists all of the certificates in a system [certificate store](https://learn.microsoft.com/en-us/windows/win32/secgloss/c-gly) and the name of the subject and all of the [certificate context](https://learn.microsoft.com/en-us/windows/win32/secgloss/c-gly) properties of each of those certificates."

nmoinvaz:
I don't think that is quite what I am looking for. It might be `CertGetCertificateChain`, however, that looks like it does more than just order the certificates from leaf->root or root->leaf. 

My problem is that TLS does guarantee that the ordering of certificates returned from the server. But the response you got, indicates that the certificate store returned from `SECPKG_ATTR_REMOTE_CERT_CONTEXT` does not guarantee any ordering/relationship. Even though (from the response) there is no guarantee, it just so happens that the ordering is in reverse for builds of Windows < Windows 11 22H2 22621, and it is in sequential order according to the TLS spec for newer builds.

alvinashcraft:
Yes, it is strange that the order has consistently changed with that recent Win 11 release. Have you raised your question/concern over at [Microsoft Q&A](https://docs.microsoft.com/answers/topics/windows-uwp.html)? There are MS Engineers and other experts who monitor those forums. I'll send another follow up email on my thread with the SSPI folks to see if they can provide any other suggestions.

nmoinvaz:
Thanks! Yes, I have posted once there before but the responses didn't help. 
https://learn.microsoft.com/en-us/answers/questions/1044186/schannel-certificate-ordering-in-windows-11-22h2-u.html

I'm trying to solve this problem for curl which we use. https://github.com/curl/curl/issues/9706. I can understand the position that I should not confer any relationship to the ordering of the certificates in the SSPI attribute's context buffer, however in practice and even in example code that I've seen in chromium and other source code on the internet people have certain expectations and perhaps erronously.

alvinashcraft:
I'm sorry you didn't get any useful info there. I'll let you know when I hear back on my email thread. Everyone on that chain is out this week.

alvinashcraft:
They'd like some more details about what you're doing with the APIs. This was the response when I asked how you can reliably enumerate the store to find the leaf and root nodes.

> What are they trying to achieve?
> If we’re talking about TLS certs, then the leaf and intermediate certs must be received from the TLS peer, and the root is discovered (or not) in the process of building the cert chain.

nmoinvaz:
We are trying to independently verify the certificate chain starting at the leaf certificate. The document associated with this PR states "Using the leaf certificate you must verify the following". The documentation doesn't state how to obtain the leaf certificate, but now states that the certificates in the context buffer can be in any order.

> If we’re talking about TLS certs, then the leaf and intermediate certs must be received from the TLS peer, and the root is discovered (or not) in the process of building the cert chain.

Yes, this is my understanding also. I had previously mentioned TLS specification because TLS RFC states **"end-entity certificate which MUST be first"**.

>    Note: Prior to TLS 1.3, "certificate_list" ordering required each
   certificate to certify the one immediately preceding it; however,
   some implementations allowed some flexibility.  Servers sometimes
   send both a current and deprecated intermediate for transitional
   purposes, and others are simply configured incorrectly, but these
   cases can nonetheless be validated properly.  For maximum
   compatibility, all implementations SHOULD be prepared to handle
   potentially extraneous certificates and arbitrary orderings from any
   TLS version, with the exception of the end-entity certificate which
   MUST be first.

https://www.rfc-editor.org/rfc/rfc8446#section-4.4.2

Even our corrected language in this document still seems kind of misleading: 

> This certificate is called the leaf certificate because it is the last certificate in the certificate chain and is farthest away from the [*root certificate*]

Older versions of Windows 11 returned the leaf certificate last (which matched with the language used in this doc) when enumerating the certificates in the certificate context using `CertEnumCertificatesInStore`. Newer versions of Windows 11 return the leaf certificate first when enumerating using `CertEnumCertificatesInStore` which matches the spec.

So how do we reliably find the leaf certificate across different versions of Windows in `CERT_CONTEXT`. Is this change related to TLS 1.3? Was TLS 1.3 introduced in Windows 11 22H2 22621?

PR #1388, opened 2022-10-24
Direct3D9: Clarify that Occlusion Queries return the number of samples, not pixels

CookiePLMonster:
As per a discussion on DirectX Discord, it's been determined with high probability that the docs for D3D9 occlusion queries mistakenly claim the query returns pixels drawn, not samples. This was deemed not to be true for the following reasons:
1. Other APIs (like D3D11 and OpenGL) refer to samples when talking about occlusion queries.
2. Tests have revealed that when multisampled rendertargets are used, the returned value is **exactly** X times higher than expected where X is the sample count of the target.

I was able to prove this via a bug in a commercial application from 2003, Colin McRae Rally 3: [THIS SUN COLORS BUG](https://www.youtube.com/watch?v=3ci8sFPCLpU) is caused exactly by the game assuming that occlusion queries return pixels, not samples, and therefore a calculated exposure value that is expected to be in `0.0f - 1.0f` range goes out of that range to e.g. `0.0f - 8.0f` when 8x MSAA is in use. Modifying the application to divide the query result by the sample count fixes the issue, which in my opinion serves as a good empirical proof:
image

### DIFF
--- a/desktop-src/direct3d9/queries.md
+++ b/desktop-src/direct3d9/queries.md
@@ -152,7 +152,7 @@ From each of the three query states, here are the [**GetData**](/windows/win32/a
 
  
 
-For example, when a query is in the issued state and the answer to the query is not available, [**GetData**](/windows/win32/api/d3d9helper/nf-d3d9helper-idirect3dquery9-getdata) returns S\_FALSE. When the resource finishes its work and the application has issued a query end, the resource transitions the query to the signaled state. From the signaled state, **GetData** returns S\_OK which means that the answer to the query is also returned in *pData*. For instance, here is the sequence of events to return the number of pixels drawn in a render sequence:
+For example, when a query is in the issued state and the answer to the query is not available, [**GetData**](/windows/win32/api/d3d9helper/nf-d3d9helper-idirect3dquery9-getdata) returns S\_FALSE. When the resource finishes its work and the application has issued a query end, the resource transitions the query to the signaled state. From the signaled state, **GetData** returns S\_OK which means that the answer to the query is also returned in *pData*. For instance, here is the sequence of events to return the number of pixels (or samples when multisampling is enabled) drawn in a render sequence:
 
 -   Create the query.
 -   Issue a begin event.
@@ -164,7 +164,7 @@ The following is the corresponding sequence of code:
 
 ```
 IDirect3DQuery9* pOcclusionQuery;
-DWORD numberOfPixelsDrawn;
+DWORD numberOfSamplesDrawn;
 
 m_pD3DDevice->CreateQuery(D3DQUERYTYPE_OCCLUSION, &pOcclusionQuery);
 
@@ -181,16 +181,19 @@ pOcclusionQuery->Issue(D3DISSUE_END);
 
 // Force the driver to execute the commands from the command buffer.
 // Empty the command buffer and wait until the GPU is idle.
-while(S_FALSE == pOcclusionQuery->GetData( &numberOfPixelsDrawn, 
+while(S_FALSE == pOcclusionQuery->GetData( &numberOfSamplesDrawn, 
                                   sizeof(DWORD), D3DGETDATA_FLUSH ))
     ;
+
+// To get the number of pixels drawn when multisampling is enabled,
+// divide numberOfSamplesDrawn by the sample count of the render target.
 ```
 
 
 
 These lines of code do several things:
 
--   Call [**GetData**](/windows/win32/api/d3d9helper/nf-d3d9helper-idirect3dquery9-getdata) to return the number of pixels drawn.
+-   Call [**GetData**](/windows/win32/api/d3d9helper/nf-d3d9helper-idirect3dquery9-getdata) to return the number of pixels/samples drawn.
 -   Specify [**D3DGETDATA\_FLUSH**](d3dgetdata-flush.md) to enable the resource to transition the query to the signaled state.
 -   Poll the query resource by calling [**GetData**](/windows/win32/api/d3d9helper/nf-d3d9helper-idirect3dquery9-getdata) from a loop. As long as **GetData** returns S\_FALSE, this means the resource has not returned the answer yet.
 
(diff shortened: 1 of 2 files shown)


CookiePLMonster:
Update: We'll be performing a regression test on this to verify whether queries behaved the same on older OSes. I am still able to reproduce this issue on a Win8.1 virtual machine, we should be able to try it on a WinXP physical machine soon. If it works as expected there, it's likely a regression in DX9 runtime.

I'll update this PR with results once we know more.

desktop-src/direct3d9/queries.md
```diff
@@ -152,7 +152,7 @@ From each of the three query states, here are the [**GetData**](/windows/win32/a
 
  
 
-For example, when a query is in the issued state and the answer to the query is not available, [**GetData**](/windows/win32/api/d3d9helper/nf-d3d9helper-idirect3dquery9-getdata) returns S\_FALSE. When the resource finishes its work and the application has issued a query end, the resource transitions the query to the signaled state. From the signaled state, **GetData** returns S\_OK which means that the answer to the query is also returned in *pData*. For instance, here is the sequence of events to return the number of pixels drawn in a render sequence:
+For example, when a query is in the issued state and the answer to the query is not available, [**GetData**](/windows/win32/api/d3d9helper/nf-d3d9helper-idirect3dquery9-getdata) returns S\_FALSE. When the resource finishes its work and the application has issued a query end, the resource transitions the query to the signaled state. From the signaled state, **GetData** returns S\_OK which means that the answer to the query is also returned in *pData*. For instance, here is the sequence of events to return the number of samples shaded in a render sequence:
```
Triang3l: "shaded" is a confusing term here as it suggests that it's related to the number of the pixel shader invocations, however, occlusion queries are purely output-merger (depth-stencil test specifically) functionality — referencing the concept of shading in this context raises at least two issues: that the depth-stencil test may be performed after pixel shading (and, if I understand correctly, in D3D9's abstract pipeline it is — just like in D3D10+ without `[earlydepthstencil]` and disregarding the pipeline statistics query), and it's not clear what "samples shaded" means with respect to shading rate (especially considering that D3D9 doesn't have sample-rate shading, thus individual samples in a pixel are never shaded separately maybe only with the exception of driver settings hacks that violate the D3D API specification).

desktop-src/direct3d9/queries.md
```diff
@@ -181,16 +181,19 @@ pOcclusionQuery->Issue(D3DISSUE_END);
 
 // Force the driver to execute the commands from the command buffer.
 // Empty the command buffer and wait until the GPU is idle.
-while(S_FALSE == pOcclusionQuery->GetData( &numberOfPixelsDrawn, 
+while(S_FALSE == pOcclusionQuery->GetData( &numberOfSamplesShaded, 
                                   sizeof(DWORD), D3DGETDATA_FLUSH ))
     ;
+
+// Divide numberOfSamplesShaded by sample count of the rendertarget
```
Triang3l: "**the** sample count", "render target" is also written as two words in Direct3D terminology. This comment is somewhat confusing overall though as it suggests that there should be code performing the division after it, but the example was truncated for some reason.

Triang3l (changes requested):

Triang3l:
I'd also suggest verifying the behavior of the Direct3D 9 reference device, though if it happens to provide the pixel count, I think it'd be safer to assume that the reference rasterizer de facto behaves incorrectly retroactively.

desktop-src/direct3d9/queries.md
CookiePLMonster: Direct3D9 seems to be inconsistent about that:
Render-target: https://learn.microsoft.com/en-us/windows/win32/api/d3d9helper/nf-d3d9helper-idirect3ddevice9-createrendertarget
Rendertarget: https://learn.microsoft.com/en-us/windows/win32/api/d3d9helper/nf-d3d9helper-idirect3ddevice9-stretchrect

CookiePLMonster: Any idea how to improve the comment? Or maybe include the division?

desktop-src/direct3d9/queries.md
CookiePLMonster: Good call, what do you think is the best terminology here to combine truthfulness and "not going into much detail"?

Triang3l: I think "drawn" that was originally here was the closest to the reality while still being abstract enough and not bringing up concepts not directly involved in this example, such as the depth-stencil buffer, as the depth-stencil test is essentially the final check (though I can't remember the exact order in the abstract D3D9 pipeline specification) that is performed to determine whether the _draw_ call should have effect on the given samples, thus if clipping, the scissor test, the alpha test, the sample mask test have passed, whether something is drawn or not in the specific location depends on the depth-stencil test, and if a sample has successfully reached the depth-stencil test, but it is disabled, the sample has been drawn in the end.

However, "drawn" alone doesn't hint at the primary purpose of occlusion queries (that is even indicated in their name) — counting how many samples were not occluded by previous draws. Therefore, I'd probably refer to both the fact that we're counting the drawn samples (I'd explicitly call them "MSAA samples" for clarity as a sample has many meanings in D3D and in graphics overall, and add something like "(when multisampling is not used, this is equal to the number of pixels drawn)") for the overall idea of what units the number is in, but also that we're counting "MSAA samples drawn that passed the depth-stencil test" so the main use case is also presented.

CookiePLMonster: I wouldn't use the term MSAA as D3D9 exclusively refers to "multisampling", but otherwise I agree.

desktop-src/direct3d9/queries.md
Triang3l: The division here would not be convenient because there are no surfaces to GetDesc of here, it's not clear which surface is actually bound at this point (only color, only depth-stencil, or both — so we don't know whether we can call GetDepthStencilSurface, and whether we can use GetRenderTarget), and the multisample setting in D3D9 is a enumeration that also contains D3DMULTISAMPLE_NONMASKABLE.

However, switching the order of the clauses in that sentence — starting it with the problem, "To get the number of pixels drawn [comma]…", rather than the solution, "Divide…" — feels to me like it changes the tone of the comment from being an explanation of what happens after that comment in the code to a suggestion for the reader what to do if they are in the specified situation, somewhat also interrupting the "flow" of reading the code doing some magic.

CookiePLMonster: Do we know how nonmaskable samples work? It always felt to me like an unwanted feature that MSDN docs say "should be used for antialiasing" but in practice I have **never** seen **anyone** use it in a commercial game.

desktop-src/direct3d9/queries.md
Triang3l: Yes, and "pixels (or multisampling samples when multisampling is enabled)" would probably be easier to understand since it starts with the simpler concept. While going from general concepts (multisampling samples) to particular cases (single-sample rendering) is natural in specifications where coverage of possible situations is the most important, for learning it's probably the opposite, otherwise we'd be teaching arithmetic in elementary school in the terms commonly used in math papers and starting from complex number matrices of arbitrary dimensionalities or something like that.

CookiePLMonster:
More empirical evidence that queries always returned samples, not pixels - a sequel to the game in question, Colin McRae Rally 04, "fixed" the flicker by clamping the sun exposure factor (`pixels_unoccluded / pixels_total`) to a `0.0f - 1.0f` range. Remove this clamp and the bug is back in an identical form:
image

This potentially provides evidence the developers added a value clamp back in 2003/2004 after this bug got noticed. I find it rather unlikely they'd go back to this code and clamp the value if they faced no issues.

desktop-src/direct3d9/queries.md
```diff
@@ -181,16 +181,19 @@ pOcclusionQuery->Issue(D3DISSUE_END);
 
 // Force the driver to execute the commands from the command buffer.
 // Empty the command buffer and wait until the GPU is idle.
-while(S_FALSE == pOcclusionQuery->GetData( &numberOfPixelsDrawn, 
+while(S_FALSE == pOcclusionQuery->GetData( &numberOfSamplesShaded,
```
Triang3l: The "shaded" word in the variable name and in the comment referencing it is a leftover from an older version of the PR.

Triang3l (changes requested):

CookiePLMonster:
@stevewhims We're still waiting for confirmation on a real WinXP machine (was unable to get MSAA on a virtual machine), but do you think this is needed or is the existing investigation enough to consider it a doc bug and correct pixels to samples unconditionally?

stevewhims:
> rtual machine, we should be able to try it on

I'm happy to consider this a doc bug, and take the PR if you can change it out of draft mode.

Incidentally, just a note about consistency.

The first replacement of "pixels" in each topic is great. It replaces "pixels" with the string "pixels (or samples when multisampling is enabled)". But after that, there's inconsistency. In some places, "pixels" is changed to "samples"; in others "pixels" is changed to "pixels (or samples)".

As I say, I do like that first replacement. That is, the change from "pixels" to "pixels (or samples when multisampling is enabled)"). But after that, how about consistently changing "pixels" to "pixels/samples"?

Thanks!
-Steve

CookiePLMonster:
That's a very good idea, thank you! I amended those changes and un-drafted the PR.

stevewhims (approved):

stevewhims:
Great, thanks for this PR (and all the investigation). And thanks for fixing that typo in there, too.

Thank you!
-Steve

PR #1390, opened 2022-10-25
Move Algorithm Constants to appropriate sections

josephlr:
The current documentation lists `BCRYPT_RC4_ALG_HANDLE` in the RNG algorithms. This is clearly an error.

This PR:
  - moves the RNG algorithms from "Block Cipher and Cipher-based Message Authentication Code Algorithms" to "Random Number Generator Algorithms"
  - moves the asymmetric algorithms from "Block Cipher and Cipher-based Message Authentication Code Algorithms" to "Asymmetric Algorithms"
  - removes the hash algorithms form "Block Cipher and Cipher-based Message Authentication Code Algorithms" as they are already listed under "Hash and Message Authentication Code (MAC) Algorithms"
  - I didn't modify or move the KDF algorithms. Maybe they should have their own section?

Other than moving lines or deleting duplicates, the content of the documentation is not modified.

Signed-off-by: Joe Richey <joerichey@google.com>


alvinashcraft (approved):
This is great. Thank you for fixing the organization of these tables. I'll create an internal work item to look at the KDF items and also discuss how long we need to keep the notes about unsupported operating systems (Vista, Windows 8, etc) here.

PR #1405, opened 2022-11-10
Remove stray lines about using DeviceIoControl

henke37:
They are clear copy and paste leftovers that don't belong in this table.


alvinashcraft (approved):
Looks good. Thanks for finding and updating this!

PR #1406, opened 2022-11-12
Added Windows Server 2019 to TLS cipher list

InvertedLlama:


alvinashcraft (approved):
This looks good, but should Windows Server 2019 be added to any of the other versions on this list? Also, the target page "tls-cipher-suites-in-windows-10-v1809.md" contains no mentions of Windows Server 2019. Perhaps that page should be updated as well.

PR #1410, opened 2022-11-17
Add API reference to portal page

SeattlePseudoGenius:
Like to the comp swapchain API pages from the comp swapchain portal page


jwmsft (approved):

jwmsft:
@SeattlePseudoGenius, thanks for the PR! Since the API reference isn't "In this section", I added a Related topics section and moved the link there instead.

PR #1414, opened 2022-11-22
Add app manifest compatibility code snippet for making it easy for most of developers (on windows 10 and 11)

pratikone:
Many users are left confused when they cannot create a WS_EX_LAYERED child window. This page has the reason, but it is not very apparent. It is easy to miss. I have added code snippets which highlights this issue so that most of the users which are developing on windows 10 and 11, will not miss this crucial information.


hickeys (approved):

hickeys:
#sign-off

PR #1415, opened 2022-11-23
Added 2 other unsupported file name to the list

Nielsbishere:
COM0 and LPT0 are also not allowed.


alvinashcraft (approved):
Looks good. Thanks for your contribution!

PR #1417, opened 2022-11-26
Correct the redistribution policy for DbgHelp

namazso:
The dll was last redistributable in the 7.1 WDK. Ever since 8 only the installer is redistributable. The current text probably wasn't updated and makes people who don't explicitly search up the license believe the dlls are still redistributable.

See the current license: http://go.microsoft.com/fwlink/?LinkId=294840

>Debugging Tools for Windows
>
>The following files may only be distributed unmodified, as a package, as part of your program:
>
>Program Files\Windows Kits\10\Debuggers\Redist\X86 Debuggers and Tools-x86_en-us.msi
>Program Files\Windows Kits\10\Debuggers\Redist\X64 Debuggers and Tools-x64_en-us.msi


namazso:
Also, I'd like to call this licensing policy change idiotic, as it only encourages shipping ancient versions, but that is not *my* problem.

Karl-Bridge-Microsoft (approved):

PR #1419, opened 2022-12-01
Fix rendering of assembly file name

Chronial:
The previous rendering was broken because of the incorrectly escaped < and > characters.


drewbatgit:
#sign-off

drewbatgit:
#sign-off

drewbatgit (approved):

PR #1425, opened 2022-12-13
Update about-keyboard-input.md

DJm00n:
Update scancodes table.

- Remove "Scan 1 break" column because we cannot see those via any of Windows API - `RAWKEYBOARD`/`WM_KEYDOWN`/`WM_SYSKEYDOWN`/`WM_KEYUP`/`WM_SYSKEYUP` are using only "Scan 1 make" codes with additional "break" flag on key up. Remove erroneous paragraph about that.
- Mention that it is "keyboard driver" who generates these scan codes currently - because HID devices don't have i8042 scan codes natively.
- Add multimedia keys to the list of scan codes.
- Remove useless `E0_` lines from the table.

Sources:
https://download.microsoft.com/download/1/6/1/161ba512-40e2-4cc9-843a-923143f3456c/scancode.doc
https://download.microsoft.com/download/1/6/1/161ba512-40e2-4cc9-843a-923143f3456c/translate.pdf
https://usb.org/document-library/hid-usage-tables-13

### DIFF
--- a/desktop-src/inputdev/about-keyboard-input.md
+++ b/desktop-src/inputdev/about-keyboard-input.md
@@ -296,8 +296,8 @@ The **Scan 1 Make** code is delivered in [**WM\_KEYDOWN**](wm-keydown.md)/[**WM\
 
 Notes:
 
-1. **System Request** key scan code is emmited on **Alt + PrintScreen** press
-2. **Break** key scan code is emmited on **Control + Pause** press
+1. *SysRq* key scan code is emmited on *Alt+Print screen* keystroke
+2. *Break* key scan code is emmited on *Control+Pause* keystroke
 3. As seen in [legacy keyboard messages](keyboard-input-notifications.md)
 4. The key is present on Brazilian keyboards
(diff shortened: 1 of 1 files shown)


DJm00n:
Need to do another pass and add HID Usages to the table and change key names in accordance to the current HID Usage Table to avoid confusion:

>The [HID Usage Tables 1.3](https://usb.org/document-library/hid-usage-tables-13) document also includes all Usage definitions as a JSON file as an attachment to the PDF. The PDF serves as the 'single' source of truth.

DJm00n:
> Need to do another pass and add HID Usages to the table and change key names in accordance to the current HID Usage Table to avoid confusion:
> 
> > The [HID Usage Tables 1.3](https://usb.org/document-library/hid-usage-tables-13) document also includes all Usage definitions as a JSON file as an attachment to the PDF. The PDF serves as the 'single' source of truth.

Done

DJm00n:
adding @matwilli for visibility.

Karl-Bridge-Microsoft (approved):

PR #1427, opened 2022-12-15
naming-a-file.txt: "Reserved" isn't a suggestion

SamB:
Use of reserved names for files is beyond "not recommended";
they simply cannot be used for normal files.

Also clarify that it doesn't matter how many dots you use.
(I checked; opening NUL.tar.gz gave me a handle to "\Device\Null".)


alvinashcraft (approved):
Looks good. Thanks for your contribution!

PR #1430, opened 2022-12-28
Update finding-the-owner-of-a-file-object-in-c--.md

amichaimazuz:
need to multiply the buffer by sizeof(wchar_t) other wise, the second call trough error


alvinashcraft (approved):
Looks good. Thanks for catching this.

PR #1431, opened 2023-01-01
Fix member types of `RESDIR` structure

squeek502:
The corrected types match [`GRPICONDIRENTRY` from here](https://learn.microsoft.com/en-us/previous-versions/ms997538(v=msdn.10)?redirectedfrom=MSDN) and [`GRPICONDIRENTRY` from here](https://devblogs.microsoft.com/oldnewthing/20120720-00/?p=7083) and empirically match what the `RC` compiler outputs.


squeek502:

jwmsft (approved):

PR #1436, opened 2023-01-16
Fix typo "installrt" -> "installer"

kruton:
This seems like a type based on "rt" being one off from "er" in a QWERTY keyboard.


alvinashcraft (approved):
Thanks for catching this typo!

PR #1439, opened 2023-01-16
Update about-introduction-to-asn-1-syntax-and-encoding.md

stesmi:
Fixed typo where "23" was written to be 23 bytes, where it's 35 to be it being hex (0x23 = 35 bytes).


alvinashcraft (approved):
Thanks for fixing this!

PR #1446, opened 2023-01-29
Update symbolic-link-effects-on-file-systems-functions.md

Norgerman:
from: https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-copyfile#:~:text=If%20the%20destination%20file%20already%20exists%20and%20is%20a%20symbolic%20link%2C%20the%20target%20of%20the%20symbolic%20link%20is%20overwritten%20by%20the%20source%20file.


alvinashcraft (approved):
Looks good. Thanks for catching this inconsistency.

PR #1457, opened 2023-02-07
pseudo-vs--true-blocking-2.md: correct nonsensical reference to 16-bit Windows

skissane:
This doesn't make sense:

> In 16 Windows environments, true blocking is not supported by the operating system

What are "16 Windows environments"? But I assume someone just forgot to say "bit", and therefore it should read:

> In 16-bit Windows environments, true blocking is not supported by the operating system

Which I believe is true, the Win16 API doesn't support "blocking" in the sense that Win32 does.

### DIFF
--- a/desktop-src/WinSock/pseudo-vs--true-blocking-2.md
+++ b/desktop-src/WinSock/pseudo-vs--true-blocking-2.md
@@ -1,21 +1,17 @@
 ---
-description: Pseudoblocking pseudo blocking operations in Windows Sockets (Winsock).
+title: Pseudo-blocking and true blocking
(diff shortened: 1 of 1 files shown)


stevewhims:

skissane:
Hi Steve, I'm surprised that a trivial one-word correction needs a CLA to be signed? But that's fine, I'll sign it:

I'm signing it on behalf of myself not my employer. Per my own understanding of my employer's IP policies, I don't believe they would claim any IP interest in minor corrections to Microsoft's product documentation. Thanks, Simon.

stevewhims (approved):

PR #1458, opened 2023-02-07
Break up numbered list into multiple lines

DavidLyhedDanielsson:
It is clearer that it is a list and is easier to follow.

~~Using only "1." makes sure the list isn't broken if items are added/removed.~~


stevewhims:
Yes, I like the technique of using `1.` for all list items. But only when the topic doesn't refer to list items by number. Here the list itself refers to items by number. In those cases I prefer to be explicit so that if anyone changes the number of items, or their sequence, there's more chance that they'll fix up the references. I removed some redundant whitespace, though.

Thanks!
-Steve

stevewhims:

DavidLyhedDanielsson:

DavidLyhedDanielsson:
> Yes, I like the technique of using `1.` for all list items. But only when the topic doesn't refer to list items by number. Here the list itself refers to items by number. In those cases I prefer to be explicit so that if anyone changes the number of items, or their sequence, there's more chance that they'll fix up the references.

I agree. Looks good to me!

stevewhims (approved):

PR #1484, opened 2023-02-11
Fixed inconsistencies in HLSL ASM formatting

Nintynuts:
I noticed that there were inconsistencies and some errors with the HLSL ASM documentation and I have tried to fix them. More could be done, but I wanted to at least do these basic things first.

In the majority of instructions, a single destination register is called `dest` so I have made any `dst` or `dst0` (where there is no `dst1`) match that.


Nintynuts:

Nintynuts:
It seems the component selection annotations are a little inconsistent too:
- ✔️ `.swizzle` any combination of xyzw (input) 
- ✔️ `.mask` any combination of xyzw (output) 
- ❌ `.write_mask` any combination of xyzw (output) - duplicate of above?
- ❓ `.select_component`/`.single_component_mask` seem to be the same but could be replaced with `.component`. I also wonder if this should never be optional, since it has to be one component?
- ❓ `.r` Usually it's on a sampler, but `sample_c` and `gather4` have it on `srcResource` (even though there's also `srcSampler`). I've only seen `.x` used which should be equivalent, but some things claim to require it and in other places it's optional. 
- ❓ `.xyzw` `store_uav_typed` claims to require a fixed swizzle on `dstUAV`, is this the only example? Why, and can the order not be changed?

Some instructions add the negative and absolute modifiers to them, and I'm not sure if that means it's only supported on those instructions, or if all inputs variables support it and only those ones have been annotated.

Also, I think ld2ms is named wrong, it's ldms from what I've seen.

Clarification on these points would be great, and I can fix those inconsistencies too.

stevewhims:
Hi, @Nintynuts, and thanks for making all these changes and submitting this PR! :)

It's an unusually large PR, with 70 files changed and many changes per file. Every change needs to be read and reviewed (it's not enough just to scan or smoke-test), which in cases like this is expensive. I've looked through every change, and they all look safe. But I'll reach out to an SME to see whether they have the bandwidth to confirm the technical accuracy of the changes that need that level of review. Those changes are hidden within changes such as changing span elements (those elements shouldn't be in the content in the first place, so that's a problem on our side) and simple lexical transforms.

I have some ideas for what we might do differently in a case like this, either next time (or this time if I can't find an SME with the bandwidth to review this PR).

First, if there's a find-replace transformation, then it's much less expensive to let me know what the transformation is. For example, "Change all dst to dest in topics inside the direct3dhlsl folder" (you could communicate that in a PR where you make just the very first change). That way I can make and publish the transform on my end without needing to look at and review every one of potentially hundreds of individual changes.

Then, for changing technical details in descriptions (such as changing valid parameter ranges), it'd be great to describe those changes (and the reason for them) in a comment in the PR. Then the SME can read and confirm those descriptions and reasons, and I can just confirm the change has been made consistently. That'll be more economical, I think. I do see a list of items in the PR (for example, about component selection annotations), which is great. But I can't tell whether you actually made those changes in this PR, or if you're asking whether some info in the docs is correct or not. In any case I'll ask the SME to look at the comments, too, and see what they think.

Perhaps some of the questions or discussion could be done on Microsoft Q&A rather than in a PR, if that's a better workflow. But we'll see what we can do here first.

Thanks again! :)
-Steve

stevewhims (changes requested):
All the changes look safe to me, but I'll request an SME to take a technical review. For info about why this is a "request changes" for now, please see my comment above.

Nintynuts:
Thanks @stevewhims, that's fair enough and I understand the need to verify it all.

My second comment here is a combination of observations and questions (particularly where there are question marks) which I wasn't confident to change before getting confirmation, so having a subject matter expert would be helpful.

Please let me know if anything specific of what I've said isn't clear.

A little more info about how I found the changes to make: I scraped all the data from the pages and made an excel spreadsheet to try and identify patterns and groupings to help me with a project, so the majority of the fixes are where there were formatting errors or inconssitencies in the data which broke my formulas.

I'm not sure if your 4th paragraph is a hypothetical example or referring to a change I've made, but if I was deliberately changing something technical I would have explained it more thoroughly in my PR description. Please let me know with a comment in the changes if you have any specific questions.

stevewhims:
Thanks, @Nintynuts. I didn't hear back from any SMEs (likely because of the effort involved in reviewing this), and I don't want this PR to get too stale.

Even though it's not ideal, I think it's best, given the circumstances, for me to take this change even without SME review. Part of that is that given the effort you've put into it, it'd be a shame not to take it. So I'll complete the PR now, *but if you could in future just describe any find/replace transforms we need to make* rather than actually making them, and keep those separate from technical edits, then that'd be great. That way, future PRs won't be so big, and I'll much more likely be able to manage them better and be able to confer with SMEs about more pointed and specific details, and not need to ask "Hey, does anyone have time to review this massive PR", which is likely to be met with "no" or silence, as in this case. :)

Thanks!
-Steve

stevewhims (approved):

PR #1489, opened 2023-02-15
Update a-shafinal.md

RatinCN:
`A_SHAFinal` follows `A_SHAUpdate`, not `MD5Update`

### DIFF
--- a/desktop-src/SecCrypto/a-shafinal.md
+++ b/desktop-src/SecCrypto/a-shafinal.md
@@ -1,5 +1,5 @@
 ---
-description: Computes the final hash of the data entered by the MD5Update function.
+description: Computes the final hash of the data entered by the A_SHAUpdate function.
 ms.assetid: A0457D26-F4E3-4ED4-B374-0AFCB6F661FB
 title: A_SHAFinal function (Sha.h)
 ms.topic: reference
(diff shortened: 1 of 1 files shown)


desktop-src/SecCrypto/a-shafinal.md
```diff
@@ -17,7 +17,7 @@ api_location:
 
 # A\_SHAFinal function
 
-Computes the final hash of the data entered by the MD5Update function.
+Computes the final hash of the data entered by the A_SHAUpdate function.
```
alvinashcraft, line 20: Thanks for this update. What do you think about mentioning that the data digest should first be initialized with a call to the `A_SHAInit` function?

RatinCN: IMO, we needn't mention `A_SHAInit` here, and there are some suggestions may helpful:
1. `A_SHAInit` and `A_SHAUpdate` mention each other, `A_SHAUpdate` and `A_SHAFinal` mention each other, the relationship amount them could be clear
2. Link all of them in `See Also` section, after all, `XxxInit` -> `XxxUpdate` -> `XxxFinal` is standard flow for MD4, MD5, SHA, ...

I can contribute them in a separate PR if you think it's can be helpful

alvinashcraft: That sounds good. I'll approve and merge this PR, and if you'd like to submit another PR for the above changes, that would be great. Thanks!

alvinashcraft (approved):
These changes look good. Thanks for your contribution!

PR #1493, opened 2023-02-16
Update a-shaupdate.md

RatinCN:
I believe `Buffer` should be a input parameter, a pointer to the data, and tested.
SAL and description are fixed in this PR.


alvinashcraft (approved):
You're correct. It should be input param. Thanks for the update!

PR #1516, opened 2023-02-28
Update a-accountexpires.md

aarontguilmette:
Updating the Remarks section per a customer service request after determining the flow of the accountExpires attribute.

### DIFF
--- a/desktop-src/ADSchema/a-accountexpires.md
+++ b/desktop-src/ADSchema/a-accountexpires.md
@@ -196,7 +196,7 @@ The date when the account expires. This value represents the number of 100-nanos
 
 The high part of this large integer corresponds to the **dwHighDateTime** member of the [**FILETIME**](/windows/desktop/api/minwinbase/ns-minwinbase-filetime) structure and the low part corresponds to the **dwLowDateTime** member of the **FILETIME** structure.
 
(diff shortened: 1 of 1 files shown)


desktop-src/ADSchema/a-accountexpires.md
```diff
@@ -196,5 +196,7 @@ The date when the account expires. This value represents the number of 100-nanos
 
 The high part of this large integer corresponds to the **dwHighDateTime** member of the [**FILETIME**](/windows/desktop/api/minwinbase/ns-minwinbase-filetime) structure and the low part corresponds to the **dwLowDateTime** member of the **FILETIME** structure.
 
+When an account is created, the account is initially set to Never Expire.  The accountExpires attribute is set to the default of 9223372036854775807, a value which corresponds the maximum value of a 64-bit signed integer. If an account is configured with an expiration date, the accountExpires value is configured to the **FILETIME** value for the expiration date. If at any point in time an account which was configured with an expiration time is set back to Never Expires, the accountExpires attribute is then set to 0.
```
alvinashcraft: Some formatting tweaks. Otherwise, looks good!

```suggestion
When an account is created, the account is initially set to Never Expire.  The **accountExpires** attribute is set to the default of `9223372036854775807`, a value which corresponds the maximum value of a 64-bit signed integer. If an account is configured with an expiration date, the **accountExpires** value is configured to the **FILETIME** value for the expiration date. If at any point in time an account which was configured with an expiration time is set back to Never Expires, the **accountExpires** attribute is then set to `0`.
```

alvinashcraft (approved):
Thanks for updating the docs!

PR #1525, opened 2023-03-04
child-sessions.md: fix list formatting

skissane:
If you look at the live docs site: https://learn.microsoft.com/en-us/windows/win32/termserv/child-sessions

You see the list of operating systems which *do not* support child sessions is all jumbled into one line:
> Child sessions are not supported on the following operating systems:
>
> Windows RT Windows Server 2012 Server Core installation option Microsoft Hyper-V Server 2012

That makes it rather hard to read, and doesn't look good either. I think it should be a bullet list instead:
> Child sessions are not supported on the following operating systems:
>
> -   Windows RT
> -   Windows Server 2012 Server Core installation option
> -   Microsoft Hyper-V Server 2012

### DIFF
--- a/desktop-src/TermServ/child-sessions.md
+++ b/desktop-src/TermServ/child-sessions.md
@@ -13,10 +13,9 @@ Beginning with Windows Server 2012 and Windows 8, Remote Desktop supports the 
 
 Child sessions are not supported on the following operating systems:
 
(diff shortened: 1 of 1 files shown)


QuinnRadich (approved):

QuinnRadich:
#signoff

PR #1531, opened 2023-03-13
Other not interoperate func should be GSS_Unwrap.

TheBetterSolution:
I think this paragraph:
Be aware that MakeSignature and VerifySignature will not interoperate with GSS_Wrap when conf_flag is set to zero, or with GSS_Wrap. 
should be changed to:
Be aware that MakeSignature and VerifySignature will not interoperate with GSS_Wrap when conf_flag is set to zero, or with GSS_Unwrap.

The last word for another function which will not interoperate with MakeSignature and VerifySignature should be GSS_Unwrap.


alvinashcraft (approved):
This change looks good. Thanks for contributing to the Windows docs!

PR #1535, opened 2023-03-26
Reinstate lost n-dashes in filesystem-functionality-comparison.md

TBBle:
As seen in the version of this file in 66168cfa76cac3f27171114d3aac3bc9df9e18ef, a bunch of values in this file are of the form "2^X–1", but a later build of the docs had replaced the n-dash with a space.

I only noticed it in this file when trying to understand what "2^64 1 bytes" meant, but it's possible this was a general import problem and should be globally addressed.


alvinashcraft (approved):
Thanks for addressing this issue in the docs. I did a quick search and didn't find any other topics with this same issue with a space instead of a dash, but I'm going to create a work item to do a more thorough search of the content.

PR #1536, opened 2023-03-27
Update styling in controlling-a-service-using-sc.md

cadamini:
The reference documentation page for sc stop does not exist for the latest version but only for server 2012, see https://learn.microsoft.com/en-us/previous-versions/windows/it-pro/windows-server-2012-r2-and-2012/cc742107(v=ws.11),
latest only lists sc config, create, delete and query, see https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/sc-config

I was looking for suitable sc documentation to link to and found https://learn.microsoft.com/en-us/windows/win32/services/controlling-a-service-using-sc. For me, this general page would benefit from an update to the same styling.

### DIFF
--- a/desktop-src/Services/controlling-a-service-using-sc.md
+++ b/desktop-src/Services/controlling-a-service-using-sc.md
@@ -12,7 +12,17 @@ The Windows SDK contains a command-line utility, Sc.exe, that can be used to con
 
 ## Syntax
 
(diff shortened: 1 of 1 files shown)


cadamini:

stevewhims (approved):

PR #1548, opened 2023-04-13
Fix typo

Lipovlan:
Adds a space between an identifier and an "or" on CNG algorithm identifier page.


alvinashcraft (approved):
Nice catch. Thanks for updating this.

PR #1553, opened 2023-04-19
Fix setsockopt params in so-keepalive.md

Gyross:
Correct the description the parameters of 'setsockopt', so that 'optlen' is an input parameter, and a value rather than pointer type. Remove 'getsockopt' from 'Syntax' section as it is not described in the 'Parameters' section.


Gyross:

SnapF21:
Ty

stevewhims (approved):

PR #1554, opened 2023-04-19
Docs: Fixed an awkward order of parameters and descriptions 

tomorrow9913:
Before revision
image
After revision
image

See also:  #1552


tomorrow9913:

stevewhims (approved):

PR #1560, opened 2023-04-21
Wrong library in pragma. Should be advapi32.lib

MarcoPeretti:
The sample does not compile as is (VS 2022, Windows 10 SDK). LookupPrivilegeValue and AdjustTokenPrivileges are exported by advapi32.lib


alvinashcraft (approved):
Thank you for fixing this code sample. Looks good.

PR #1562, opened 2023-04-23
Fix default value of Direct2D's 2D Affine Transform effect's sharpness property

rickbrew:
The default value for the `D2D1_2DAFFINETRANSFORM_PROP_SHARPNESS` property is actually 0, not 1. You can verify this by instantiating the effect and retrieving the property value at runtime. 

0.0f is also the documented value for the sharpness property on the Scale effect ( https://learn.microsoft.com/en-us/windows/win32/direct2d/high-quality-scale ), so it makes sense that these two should match.


stevewhims (approved):
Thank you! :)

PR #1567, opened 2023-04-29
Change metadata to see exponent operation instead of multiplication

LuridHound:


LuridHound:
Suppose it's typo, that is written 2*x instead of 2^x.

stevewhims (approved):
Thank you! :)

PR #1588, opened 2023-05-17
Update toc.yml

DJm00n:
fix wrong links

### DIFF
--- a/desktop-src/menurc/toc.yml
+++ b/desktop-src/menurc/toc.yml
@@ -38,11 +38,11 @@
         - name: "EnumResLangProc"
           href: /previous-versions/windows/desktop/legacy/ms648033(v=vs.85)
         - name: "EnumResNameProc"
-          href: /previous-versions/windows/desktop/legacy/ms648034(v=vs.85)
+          href: /windows/desktop/api/libloaderapi/nc-libloaderapi-enumresnameprocw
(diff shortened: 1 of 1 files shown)


GrantMeStrength:
Drew, Jim  

This PR proposes a fix to the TOC, as it is pointing to a topic in the archive, but another identical looking topic still exists.
Is this as expected? Are there more?

Original
https://learn.microsoft.com/en-us/previous-versions/windows/desktop/legacy/ms648034(v=vs.85)

Proposed correction:
https://learn.microsoft.com/en-us/windows/win32/api/libloaderapi/nc-libloaderapi-enumresnameprocw

The links in the archived version e.g. EnumResourceNamesEx jumps to topics back outside the archive.

drewbatgit (approved):

drewbatgit:
@GrantMeStrength The archive page was grabbed by the "catch-all" archive pass which grabbed all topics from MSDN that hadn't already been migrated and put them in this archive. So, this is different that the pages we archived deliberately. I suspect this one was grabbed because the page was split into A and W versions.  There probably are more instances where this occurred. Let me know if you want me to investigate/fix this.

The links from the archive page jump back outside the archive because they are msdn links that are redirected to the live pages. 

 I've approved this change and will go ahead and complete the PR.

PR #1589, opened 2023-05-17
Update -wic-about-windows-imaging-codec.md

Alcaro:
There's also mention of a Digital Negative (DNG) format, but it's tagged prerelease, so I'm not sure if it should be mentioned in this list.


stevewhims (approved):

stevewhims:
We can add DNG after it's GA.

PR #1601, opened 2023-05-21
Update image-stride.md

MichaelaMWilson:
When setting the source pixels to the start of scan line 0, they should be set from the source image's scan line, and vice versa for the destination pixels/image.


drewbatgit (approved):

drewbatgit:
Thanks for catching this.

PR #1608, opened 2023-05-25
Fix: Update User Themes Path in themesfileformat-overview.md

jimbrig:
In the documentation for the [Theme File Format - Win32 apps](https://learn.microsoft.com/en-us/windows/win32/controls/themesfileformat-overview#example-of-a-theme-file) the section 
[Installing Theme Files](https://learn.microsoft.com/en-us/windows/win32/controls/themesfileformat-overview#installing-theme-files) currently states:

> The system default theme files are located in %WinDir%\Resources\Themes. The user theme files are stored in %WinDir%\Users\<username>\AppData\Local\Microsoft\Windows\Themes.

which is not correct with regards to the *User* theme file location.

I changed it to reference `%LOCALAPPDATA%\\Microsoft\\Windows\\Themes` instead of `%WinDir%\Users\<username>\AppData\Local\Microsoft\Windows\Themes`.

### DIFF
--- a/desktop-src/Controls/themesfileformat-overview.md
+++ b/desktop-src/Controls/themesfileformat-overview.md
@@ -664,7 +664,7 @@ Transparency=1
 
 ## Installing Theme Files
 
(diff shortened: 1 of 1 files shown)


jimbrig:
@jwmsft let me know if you need anything more..

jwmsft (approved):

jwmsft:
Thanks for the update.

jimbrig:
> Thanks for the update.

Thanks for the review. Quick fix but worth it for end readers

PR #1650, opened 2023-07-25
Add documentation for ShareHandler

amittere:
Adding documentation for how cloud file providers can support handling the "Share" command in File Explorer when invoked on a cloud file under their sync root. This functionality has been supported since Windows 11 version 21H2.


alvinashcraft (approved):
Looks great. Thanks for improving/updating these docs.

PR #1655, opened 2023-07-30
Fix the example of appending one file to another

ChrisDenton:
The example in [appending-one-file-to-another-file.md](https://github.com/MicrosoftDocs/win32/blob/8761e854c2fa1ef1cb0b7a8ace5f481400f0f5b3/desktop-src/FileIO/appending-one-file-to-another-file.md) does not actually lock the file and this goes unnoticed because it also doesn't check if locking succeeds. [`LockFile`](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-lockfile) is documented as requiring either `FILE_GENERIC_READ` or `FILE_GENERIC_WRITE` but the example only had `FILE_APPEND_DATA`. This is fixed by using `FILE_APPEND_DATA | FILE_GENERIC_READ`. You only actually need to add `FILE_READ_DATA` but I thought it best to go with the documented requirement. Tbh, I'm not entirely sure why this is necessary but it is.

Also the example was inconsistent in the case used for the file name. Modern Windows filesystems, in some configurations, can be case sensitive so it's wise to be consistent.

### DIFF
--- a/desktop-src/FileIO/appending-one-file-to-another-file.md
+++ b/desktop-src/FileIO/appending-one-file-to-another-file.md
@@ -45,7 +45,7 @@ void main()
 
   if (hFile == INVALID_HANDLE_VALUE)
   {
-     printf("Could not open One.txt."); 
+     printf("Could not open one.txt."); 
      return;
   }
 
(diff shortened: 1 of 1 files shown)


alvinashcraft (approved):
These changes look good. Thanks for improving the example code in this topic.

PR #1657, opened 2023-08-02
Fix check for GetTokenInformation failure

lb90:


alvinashcraft (approved):
Good catch. Thanks for improving the docs!

PR #1661, opened 2023-08-16
Update sparse-files-and-disk-quota.md

ben-thul:
Change gendered pronouns in example to gender neutral.


alvinashcraft (approved):
Thank you for finding and fixing this. Looks good.

PR #1673, opened 2023-09-03
Fix formatting of HLSL variable precise modifier

berniegp:
This fixes the formatting of the list and code blocks embedded in a Markdown table.


berniegp:
Actually, looking at the blame history, I see that perhaps some part of that last `<li>` should be outside the `<ul>`. I also don't know the syntax to mark the `<pre>` block as HLSL code.
https://github.com/MicrosoftDocs/win32/blame/25211012b002a7d1303e438277373d7faf958a68/desktop-src/direct3dhlsl/dx-graphics-hlsl-variable-syntax.md

stevewhims:
There's now an effort to keep HTML out of our markdown files, and remove HTML from them. So a table is probably not the best design for the presentation of this info. I'll take this change, but please add any more HTML to .md files. Thank you! :)

stevewhims (approved):

PR #1676, opened 2023-09-07
WIP: Update virtual-key-codes.md

DJm00n:
TODO:
Be in sync with
https://learn.microsoft.com/en-us/dotnet/api/system.windows.forms.keys
https://github.com/dotnet/dotnet-api-docs/blob/main/xml/System.Windows.Forms/Keys.xml
https://github.com/dotnet/winforms/blob/main/src/System.Windows.Forms/src/System/Windows/Forms/Keys.cs
and 
https://learn.microsoft.com/en-us/uwp/api/windows.system.virtualkey
https://github.com/MicrosoftDocs/winrt-api/blob/docs/windows.system/virtualkey.md


Karl-Bridge-Microsoft (approved):

qkdreyer:
The content is now badly formatted : https://learn.microsoft.com/en-us/windows/win32/inputdev/virtual-key-codes

image

Karl-Bridge-Microsoft:
THanks @qkdreyer! This should be fixed now.

PR #1680, opened 2023-09-08
Fix parameter type for CalculateThreadgroups

rickbrew:
Looks like the first parameter was copy-pasted from the `SetComputeInfo` paragraph above. It should be `const D2D1_RECT_L *`, not `ID2D1ComputeInfo *`.

Correct parameters are listed at https://learn.microsoft.com/en-us/windows/win32/api/d2d1effectauthor/nf-d2d1effectauthor-id2d1computetransform-calculatethreadgroups


stevewhims (approved):

fdwr:
Thanks Rick.

PR #1683, opened 2023-09-13
Update addprintprocessor.md

tahseenbt:
According to the following twitter thread, it is possible (and is being done in the wild) to install a print processor outside the dedicated system print-processor directory:

https://twitter.com/ESETresearch/status/1506904404225630210


drewbatgit:
@tahseenbt Sorry for the delayed response. I consulted with the feature team and they have asked me to reject your change because "we are doing an explicit check to make sure the path to the print processor being added is under the print processor directory of the Spooler." They plan do some investigation on the report in the linked tweet. Thanks for raising this issue.

PR #1688, opened 2023-09-15
Fix provider enumeration sample

heaths:
The code is incorrect and missing a lib import.


alvinashcraft (approved):
Looks good. Thanks for the fix!

PR #1690, opened 2023-09-18
Improve the dwID-Index mapping logic

Ironyho:
Improve the dwID-Index mapping logic to adapt to the situation where dwID keeps increasing on some touch devices.


Ironyho:
@Karl-Bridge-Microsoft

Karl-Bridge-Microsoft (approved):

PR #1693, opened 2023-09-20
Update considerations-for-active-directory-domain-services-backup.md

jonny190:
Same as in commit a2a16ee clarifying that 3rd party applications my function differently


alvinashcraft (approved):
Looks good. Thanks for updating this doc as well.

PR #1695, opened 2023-09-21
Update virtual-key-codes.md

DJm00n:
Remove mention of old `122-key Model M` VK codes. These are not used for about 30 years now...


Karl-Bridge-Microsoft:
Thanks @DJm00n. I'm closing without merging as these codes, however obsolete, are still defined in Windows source code.

PR #1706, opened 2023-09-27
Fix VFW_S_NO_STOP_TIME value

Alcaro:
https://godbolt.org/z/PnxWGqGr6 says VFW_S_NO_STOP_TIME isn't 0x80040270


stevewhims (approved):
Yes, confirming your value is the one in the header. Thanks!

PR #1713, opened 2023-10-15

shalala66:
Added "const qualifiers" in 3 places.

"string" is also called as "character constants", "string constant" and "anonymous string" in C++. This is because a "string"  in the C++ standard has the "const qualifier".
But the situation in C is slightly different. Since the "const keyword" did not exist in C until the "ANSI C 1989" standard,  "string literals" dont have a "const qualifier" for such historical reasons. Hence "non-const string" in C is evaluated as  "undefined behavior".
"string literals", i.e. the elements of the "envVarStrings" array are located in memory in the ".rodata section" of the "data segment", but the "pointers" of this array even though are in the ".rwdata section", they "point" to the "first  character addresses", i.e. "base addresses" of "strings" located in the "read-only section" and store these addresses. That is why must be added the "const qualifier" that indicating they are "pointing" to "const character".  Although the "printError" function has a "non-const" parameter - void printError(TCHAR* msg); - , the program string attempts to pass a "const" argument to this function. - "printError( TEXT("GetComputerName") );" To eliminate the problem the "printError" function parameter must have the "const qualifier".

### DIFF
--- a/desktop-src/SysInfo/getting-system-information.md
+++ b/desktop-src/SysInfo/getting-system-information.md
@@ -16,7 +16,7 @@ The following example uses the [**GetComputerName**](/windows/desktop/api/Winbas
 #include <tchar.h>
 #include <stdio.h>
 
-TCHAR* envVarStrings[] =
+const TCHAR* envVarStrings[] =
 {
   TEXT("OS         = %OS%"),
   TEXT("PATH       = %PATH%"),
@@ -25,7 +25,7 @@ TCHAR* envVarStrings[] =
 };
 #define  ENV_VAR_STRING_COUNT  (sizeof(envVarStrings)/sizeof(TCHAR*))
 #define INFO_BUFFER_SIZE 32767
-void printError( TCHAR* msg );
+void printError(const TCHAR* msg );
 
 void main( )
 {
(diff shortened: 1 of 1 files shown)


stevewhims (approved):

stevewhims:

Thanks again!
-Steve

shalala66:
Yes, You are right. It will be done. I already gathered them together for some PRs in the future. I have looked at official page where is full program codes loacted and noticed that there are some shortcomings in the code sections again. Maybe it is related with merge conflicts as you mentioned. Since several repetitions of same lines present in the program code, that is why I fixed all without any shortages together. Please notify me if you can change them yourself or I can make new PR. I note all of them here too.

These lines should be removed:
1. "TCHAR infoBuf[INFO_BUFFER_SIZE];" line in main function
2. 2nd "TCHAR sysMsg[MAX_PATH] = {'\0'};" line in printError function

These lines should be changed:
1. parameter 256 to MAX_PATH in FormatMessage function

These lines should be added:
1. TCHAR* p = sysMsg; line after "TCHAR sysMsg[MAX_PATH] = {'\0'};" in printError function
2. "p = nullptr;" to the last line of printError function

Thanks in advance)

PR #1716, opened 2023-10-15

shalala66:
Let's take a look at the "declaration" in the input part of the "printError()" function. After editing, the "uninitialized variable, array and wild pointer" are gone, the size of the "array" is replaced by the "MAX_PATH (256->MAX_PATH)" macro, which defines the maximum length value (currently 260) and is more efficient in my opinion. Which value is also supported by "lpBuffer and nSize - 64K bytes = 64000 bytes" parameters of "FormatMessage()" function.


stevewhims (approved):

stevewhims:

Thanks again!
-Steve

PR #1717, opened 2023-10-15

shalala66:
Even if the sysMsg array would not be modified, this part is highly suboptimal and can be performed more optimally with  a single loop. You can test it without sysMsg array modification to be sure this loop is very non optimal.  In the first "while loop", all "control characters" - [0 - 32) - except for "9 - \t (horizontal tab)" - "p - pointer" -  are "pre- increment" and the following two ending with different characters depending on the version. :

// after TCHAR sysMsg[MAX_PATH] = { '\0' }; and while( ( *p > 31 ) || ( *p == 9 ) ) version p: 000000A131EEEC04 *p: 13 <---\

NOTE: index sysMsg[50]: \r = 13 (cycle ends with this character)

// after TCHAR sysMsg[MAX_PATH] = { '\0' }; and 1st modification of the while(((*p > 31) && (*p != 46)) || (*p == 9)) version p: 000000FEDA0EECC2 *p: 46 <---\ after "while" until entering "do-while" loop

NOTE: index sysMsg[49]: . = 46 (cycle ends with this character)

Then, a new second do-while loop kicks in, which I think is more readable in the modified version -
- while ((p >= sysMsg) && ((*p == '.') || (*p < 33))); --> while ((p >= sysMsg) && ((*p == 46) || (*p < 33))); The most interesting is the line of code where "post-decrement + indirection" operations are executed together: *p-- = 0; The sequence of operations performed by this code snippet can be found from this official source - https://en.cppreference.com/w/c/language/operator_precedence -
- we can learn. The post-decrement operation has a higher priority than the indirection (dereference). Moreover, since it is also true from the associativity point of view, we can group this piece of code like this: *(p--) = 0; The value of the expression "post-decrement" is the value of p before the "decrement" operation, but the value of p is also "decremented" at the same time. Actually "operator-- overloading" occurs during "post-decrement" - T T::operator--(int); ---> T operator--(T& a, int); - and in the first step, it creates a copy of the received p, then "decrements" the value of p, and finally from the "decrement" process in the first step returns the previously created copy. Then *(p--) "indirection (dereference)" is applied to these operations:

// after TCHAR sysMsg[MAX_PATH] = { '\0' }; and while( ( *p > 31 ) || ( *p == 9 ) ) version p: 000000A131EEEC04 *p: 13 <---\ when entering "do-while" loop, *(p--) = 0; before the line p: 000000A131EEEC02 *p: 46 <---\ "do-while" loop, after line - *(p--) = 0; p: 000000A131EEEC00 *p: 1099

// after TCHAR sysMsg[MAX_PATH] = { '\0' }; and 1st modification of the while(((*p > 31) && (*p != 46)) || (*p == 9)) version p: 000000FEDA0EECC2 *p: 46 <---\ when entering "do-while" loop, *(p--) = 0; before the line p: 000000FEDA0EECC0 *p: 1099 <---\ "do-while" loop, *(p--) = 0; after the line

In general, it is very inconvenient to run 2 loops to perform the cleaning process and it is can be adjusted with a single loop. For this, exactly these code parts should be "refactored".

### DIFF
--- a/desktop-src/SysInfo/getting-system-information.md
+++ b/desktop-src/SysInfo/getting-system-information.md
@@ -75,7 +75,7 @@ void main( )
 void printError( TCHAR* msg )
 {
   DWORD eNum;
-  TCHAR sysMsg[256];
+  TCHAR sysMsg[MAX_PATH] = {'\0'};      //  it can remain as it is: TCHAR sysMsg[256];
   TCHAR* p;
 
   eNum = GetLastError( );
@@ -87,10 +87,14 @@ void printError( TCHAR* msg )
 
   // Trim the end of the line and terminate it with a null
   p = sysMsg;
-  while( ( *p > 31 ) || ( *p == 9 ) )
-    ++p;
-  do { *p-- = 0; } while( ( p >= sysMsg ) &&
-                          ( ( *p == '.' ) || ( *p < 33 ) ) );
+  while (*p++)
+  {
+      if ((*p != 9 && *p < 32) || *p == 46)
+      {
+          *p = 0;
+          break;
+      }
+  }	
 
   // Display the message
   _tprintf( TEXT("\n\t%s failed with error %d (%s)"), msg, eNum, sysMsg );


stevewhims (approved):

stevewhims:
These look like good changes, but if they'd been done in one shot that'd be ideal. This way, each PR needs merge conflicts resolved due to the changes coming in piecemeal, and each affecting the same parts of the code.

Thanks! :)
-Steve

stevewhims:

Thanks again!
-Steve

shalala66:
Yes, You are right. It will be done. I already gathered them together for some PRs in the future. I have looked at official page where is full program codes located and noticed that there are some shortcomings in the code sections again. Maybe it is related with merge conflicts as you mentioned. Since several repetitions of same lines present in the program code, that is why I fixed all without any shortages together. Please notify me if you can change them yourself or I can make new PR. I note all of them here too. 

These lines should be removed: 
1. "TCHAR  infoBuf[INFO_BUFFER_SIZE];" line in main function
2. 2nd "TCHAR sysMsg[MAX_PATH] = {'\0'};" line in printError function

These lines should be changed: 
1. parameter 256 to MAX_PATH in FormatMessage function
2. NULL parameters to nullptr in FormatMessage function

These lines should be added:
1. TCHAR* p = sysMsg; line after "TCHAR sysMsg[MAX_PATH] = {'\0'};" in printError function
2. "p = nullptr;" to the last line of printError function
3. double colon before the all function names

Thanks in advance)

PR #1718, opened 2023-10-16
[stevewhims: recommend closing] Update traversing-the-module-list.md

shalala66:
1) 
Added "const qualifiers" in 2 places.
"string" is also called as "character constants", "string constant" and "anonymous string" in C++. This is because a  "string" in the C++ standard has the "const qualifier". But the situation in C is slightly different. Since the  "const keyword" did not exist in C until the "ANSI C 1989" standard, "string literals" dont have a "const qualifier"  for such historical reasons. Hence "non-const string" in C is evaluated as "undefined behavior". Although the "printError" function has a "non-const" parameter - void printError(TCHAR* msg); -, the program string attempts to pass a "const" argument to this function. For example: "printError( TEXT("Module32First") );" To eliminate the problem the "printError" function parameter must have the "const qualifier".
Also added double colon before the some function names to emphasize that they are part of Windows API.

2)
Let's take a look at the "declaration" in the input part of the "printError()" function. After editing, the "uninitialized variable, array and wild pointer" are gone, the size of the "array" is replaced by the "MAX_PATH (256->MAX_PATH)" macro, which defines the maximum length value (currently 260) and is more efficient on my opinion. Which value is also supported by "lpBuffer and nSize - 64K bytes = 64000 bytes" parameters of  "FormatMessage()" function.

3)
Even if the sysMsg array would not be modified, this part is highly suboptimal and can be performed more optimally with  a single loop. You can test it without sysMsg array modification to be sure this loop is very non optimal. In the first "while loop", all "control characters" - [0 - 32) - except for "9 - \t (horizontal tab)" - "p - pointer" - are "pre- increment" and the following two ending with different characters depending on the version:

// after TCHAR sysMsg[MAX_PATH] = { '\0' }; and while( ( *p > 31 ) || ( *p == 9 ) ) version p: 000000A131EEEC04 *p: 13 <---\

NOTE: index sysMsg[50]: \r = 13 (cycle ends with this character)

// after TCHAR sysMsg[MAX_PATH] = { '\0' }; and 1st modification of the while(((*p > 31) && (*p != 46)) || (*p == 9)) version p: 000000FEDA0EECC2 *p: 46 <---\ after "while" until entering "do-while" loop

NOTE: index sysMsg[49]: . = 46 (cycle ends with this character)

Then, a new second do-while loop kicks in, which I think is more readable in the modified version -
- while ((p >= sysMsg) && ((*p == '.') || (*p < 33))); --> while ((p >= sysMsg) && ((*p == 46) || (*p < 33))); The most interesting is the line of code where "post-decrement + indirection" operations are executed together: *p-- = 0; The sequence of operations performed by this code snippet can be found from this official source - https://en.cppreference.com/w/c/language/operator_precedence -
- we can learn. The post-decrement operation has a higher priority than the indirection (dereference). Moreover, since it is also true from the associativity point of view, we can group this piece of code like this: *(p--) = 0;  The value of the expression "post-decrement" is the value of p before the "decrement" operation, but the value of p is also "decremented" at the same time.  Actually "operator-- overloading" occurs during "post-decrement" - T T::operator--(int); ---> T operator--(T& a, int); - and in the first step, it creates a copy of the received p, then "decrements" the value of p, and finally from the "decrement" process in the first step returns the previously created copy. Then *(p--) "indirection (dereference)" is applied to these operations:

// after TCHAR sysMsg[MAX_PATH] = { '\0' }; and while( ( *p > 31 ) || ( *p == 9 ) ) version p: 000000A131EEEC04 *p: 13 <---\ when entering "do-while" loop, *(p--) = 0; before the line p: 000000A131EEEC02 *p: 46 <---\ "do-while" loop, after line -  *(p--) = 0;
p: 000000A131EEEC00 *p: 1099

// after TCHAR sysMsg[MAX_PATH] = { '\0' }; and 1st modification of the while(((*p > 31) && (*p != 46)) || (*p == 9)) version p: 000000FEDA0EECC2 *p: 46 <---\ when entering "do-while" loop, *(p--) = 0; before the line p: 000000FEDA0EECC0 *p: 1099 <---\ "do-while" loop, after line -*(p--) = 0; 

In general, it is very inconvenient to run 2 loops to perform the cleaning process and it can be adjusted through a single loop. For this, exactly these code parts should be "refactored".

4)
These lines were changed:
NULL parameters to nullptr in FormatMessage function


shalala66:

shalala66:
@Karl-Bridge-Microsoft can You give an attention to these commits?

Karl-Bridge-Microsoft:
@shalala66 - thanks for the contributions! As these are fairly extensive revisions to code samples, they will take some time to review. The feature team will likely need to get involved for that review, which can add to delays in processing.

shalala66:
@Karl-Bridge-Microsoft - on my opinion the checking should have taken very little time. Because everything in commits are written as possible as obviously. Plus, You could verify it just by executing the code snippet and compare the results. In that case the consequences are straight forward.

shalala66:
@Karl-Bridge-Microsoft - It's already approximately 2 month that you did not paying necessary attention to this and other 2 significant pr's for Microsoft documentation. It's not about some commas, dots, other punctuation marks and etc. in text of documents. All commits including this one are about code level errors and non-optimal solutions. This code snippet has the potential error. I wrote about it full enough and this program code contains non-optimized algorithm in "printError" function. As I mentioned in above comments I actually explained entire problem in code fragments, but despite this no any attentiveness from your side during long period (+-2 month).

shalala66:
@stevewhims, @TylerMSFT 
        I'm sorry for tagging and disturbing you all here, but I've not seen any another way in such as situation,- I only know you two. Why doesn't this person give the even the slightest attention to these comments and more importantly to commits (3 commits together)? In this case why this person is here? Is it not his work? Please, if you have a time, spend on these commits. Because the responsible person does not give an any attention for solving any of them. I've already encountered several times in demo codes, mini programs, etc. very serious errors and even non-working code snippets in the official Microsoft platform for which some of you are responsible. It's not problem. Problem is that people like me spend a lot of effort who's interested in solve some errors to enhance quality of examples in official platform for other learners and clarify are ignored by the lazybones like the @Karl-Bridge-Microsoft. This is very unmotivated activity from Microsoft. I still don't understand why such people work in these positions in company as Microsoft.

stevewhims:
@TylerMSFT: Just a heads up that I'm taking a look at this now. Just letting you know so that we don't both work on it at the same time! :)

I'll report back as soon as I've processed the issue.

stevewhims:
Thanks for this PR, @shalala66! In response, I've reviewed the topic, and I've tried out the code, and I've read all the points you make in the PR.

In reviewing the topic, I saw that the code example didn't compile (as published at the time). So I've published a fix for that. But please read on for reasons why I'm recommending that this particular PR be closed.

Now that the code example builds, runs, and functions as expected, I think that's the highest priority update that was needed for this topic. The other changes in this PR are about improving efficiency, and using modern and C++-specific idioms (for example, updating `NULL` to `nullptr`, and things like that). And I don't think those changes meet the bar, nor are great return-on-investment for this topic (which has very low page views).

I know that you put a lot of effort (and expertise) into this PR, and I appreciate that. I'm always tempted myself to make any segment of code more efficient. And that's appropriate for production code. But the focus of this topic is really only to show how to use the APIs. Showing how to use the APIs *in a segment of code that's as efficient as possible* is a worthy goal. But it's not a high priority one (considering the other work we have to do). One of your points is about improving code readability. And that's important. But I just don't think that this topic needs this much time and effort at this stage.

So I'd recommending that this particular PR be closed now. Other folks might disagree, and that's fine.

Thanks!
-Steve

stevewhims (changes requested):
I recommend closing. See my comment in the thread. If you disagree, feel free to override, and take the PR.

Thanks!
-Steve

shalala66:
@stevewhims, I didn't think You would solve the problem via this way. But this contribution is related with your name in official site. It's wrong to do this on my opinion. That's not why I spent so much effort. In this case add my name to contributor lists in 2 places at least. This is my right. Okay?

shalala66:
@stevewhims, I'm waiting for your comeback about the correction of mistake that you have provided as yours contribution instead of mine. Please, correct this mistake and add my name to contributors list in at least 2 places.

stevewhims:
Hi, @shalala66. I confess (and apologize) that I hadn't given any thought to attribution (getting credit). I'm grateful to you for bringing this documentation to my attention; that's what led me to discover (and fix) the build issue. And thanks for mentioning attribution just now. I can see that that's important to you, so I'll definitely bear that in mind in future. But this GitHub issue is still open. Until it's either merged or closed, it's probably premature to worry too much about attribution.

shalala66:
Hi, @stevewhims. 
Excuse me again to attract you to this problem in this repo. I have 2 interesting questions for you. 
First, closing this will resolve the present problem? If yes, I'm ready to close. No problem. 
Second, why this guy - @Karl-Bridge-Microsoft - does not merge this pull rrequest? Is he alive? Or maybe he has some personal problems? 
I got an attention that here I talk with you whom does not related to these commits and you wanna try to solve current trouble to help, but this guy behaves ownself like if he has nothing to do with this topic. I really don't understand what is going on?

stevewhims:
I can let you know that merging the PR _will_ apply the PR's changes to the topic. Closing the PR will _not_ apply the PR's changes to the topic. Here, "closing" really means "abandoning" or "canceling".

shalala66:
Is it not injustice? Okay, let it be. In this case I have no other way except asking you to help me or providing some works in this side to add my name to contributors list in 2 places at least.

Karl-Bridge-Microsoft:
Hi @shalala66 - thanks for your suggestions and your patience. As @stevewhims states, he's fixed the code snippets so they build cleanly and published those updates. We both agree that is sufficient for these ref topics. Please note that we have limited resources, a finite amount of time, and thousands of topics to maintain, which requires us to prioritize based on significance, schedule, and workload.

As it stands, we've noted your suggestions and will circle back should time permit. 

Closing PR as topics have been updated in a private PR.

shalala66:
That is good news. @stevewhims, @Karl-Bridge-Microsoft, I just want to you is to add my name to contributor lists in the places where the changes were attempted. That's it.

PR #1726, opened 2023-10-20
fix missing IDC_PEN

expikr:


DJm00n:
`IDC_PEN` is not defined in `WinUser.h`.
See https://github.com/MicrosoftDocs/win32/pull/1738

Karl-Bridge-Microsoft (approved):

PR #1731, opened 2023-10-23
Add reserved COM/LPT names with CP-1252 superscript digits.

sashabu:
Windows recognises the 8-bit ISO/IEC 8859-1 superscript digits ¹, ², and ³ as digits and treats them as valid parts of COM# and LPT# device names, making them reserved in every directory. As far as I can tell, this is the case on all NT versions.

For example, `echo test > COM¹` fails to create a file.

Tangentially, COM0 and LPT0 are in practice treated as normal file names, contrary to documentation; I have not changed this as it's not clear whether they're still reserved for future use or just in case.


alvinashcraft (approved):
Thanks for the update!

PR #1734, opened 2023-10-30
Return value error in documentation.

AshenvaleZX:
The return value of ObjectToWorld3x4()/WorldToObject3x4() should be float3x4 and the return value of ObjectToWorld4x3()/WorldToObject4x3() should be float4x3.


AshenvaleZX:

stevewhims (approved):

PR #1738, opened 2023-11-10
Add table with additional cursors

DJm00n:
These IDs are also available and used in WPF: https://learn.microsoft.com/dotnet/api/system.windows.input.cursortype

https://github.com/dotnet/wpf/blob/main/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/Cursor.cs#L373-L385

`OCR_RDR*` cursors are internally used in Win32 in "reader mode" that is activated with [EnterReaderModeHelper](https://learn.microsoft.com/en-us/windows/win32/inputdev/enterreadermodehelper).


DJm00n:
@Karl-Bridge-Microsoft can you please take a look?

Karl-Bridge-Microsoft (approved):

PR #1760, opened 2023-12-09
WM_POWERBROADCAST: Add RegisterSuspendResumeNotification link

forderud:
Applications need to call this function first in order to enable delivery of `PBT_APMSUSPEND`, `PBT_APMRESUMEAUTOMATIC` & `PBT_APMRESUMESUSPEND` events. It's therefore natural to add a link to `RegisterSuspendResumeNotification` in the `WM_POWERBROADCAST` docs.

The documentation should ideally also be updated to mention that `PBT_APMSUSPEND`, `PBT_APMRESUMEAUTOMATIC` & `PBT_APMRESUMESUSPEND` events are not delivered by default. This is at least the behavior I'm observing when testing on Win10 and Win11. However, I would prefer if someone from Microsoft could do this, since they have a better picture of the intended behavior.

### DIFF
--- a/desktop-src/Power/wm-powerbroadcast.md
+++ b/desktop-src/Power/wm-powerbroadcast.md
@@ -82,3 +82,4 @@ The following messages are not supported on any of the operating systems specifi
 
 * [WM\_POWERBROADCAST Messages](wm-powerbroadcast-messages.md)
 * [Power Management Messages](power-management-messages.md)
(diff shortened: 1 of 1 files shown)


Karl-Bridge-Microsoft (approved):
Thanks @forderud! I'll approve this PR and will send out queries about the other detail you point out. If I hear back from a current owner or SME, I'll update the topic as required.

forderud:
Thanks a lot @Karl-Bridge-Microsoft

PR #1764, opened 2023-12-19
Fix incorrect cleanup order and error handling for global atom table …

binki:
…entries

The code example checks if `GlobalAddAtom()` fails and appears to intend to only call `GlobalDeleteAtom()` on atoms which were successfully allocated. However, the logic was flawed and would try to delete an atom which had failed to be allocated. This change corrects the error handling.


alvinashcraft (approved):
Nice catch. Thanks for your contribution!

PR #1768, opened 2023-12-21
Update dpi-and-device-independent-pixels.md

RatinCN:
Both of  the value `GetDeviceCaps` returns and `USER_DEFAULT_SCREEN_DPI` are `int`, integer division result is truncated.

If at 144 DPI, the division result will be truncated to 1.0f, but expect 1.5f.


RatinCN:
Any progress? @QuinnRadich @GrantMeStrength @Karl-Bridge-Microsoft

Karl-Bridge-Microsoft (approved):

PR #1772, opened 2024-01-02
Correct and improve `power-setting-guids.md` (and related markdown)

Enyium:
[`power-setting-guids.md`](https://learn.microsoft.com/en-us/windows/win32/power/power-setting-guids) is utter incorrect chaos.

Occurrences of different HTML tags:

- `<dl>`: 24
- `</dl>`: 23
- `<dt>`: 62
- `</dt>`: 51

The "Note" box is under the wrong GUID.

This and other Markdown files also prevent corrent rendering in some cases and pass raw Markdown through to the output. (Search [here](https://learn.microsoft.com/en-us/windows/win32/power/power-setting-guids) for "LIDSWITCH", or [here](https://learn.microsoft.com/en-us/windows/win32/power/pbt-apmpowerstatuschange) and [here](https://learn.microsoft.com/en-us/windows/win32/power/wm-powerbroadcast) for "uMsg", e.g.)

---

There are also various inconsistencies and imperfections:

- *Some* letters in GUIDs lowercase,
- bad document flow (sections not quickly distinguishable),
- missing references to existing identifiers (only bare numbers),
- missing data type.

---

Unresolved questions regarding `power-setting-guids.md`:

- `GUID_GLOBAL_USER_PRESENCE` and `GUID_SESSION_USER_PRESENCE`: What's with the other values from the `USER_ACTIVITY_PRESENCE` enum? Are they also possible values? Probably only `PowerUserNotPresent`.
- `GUID_LIDSWITCH_STATE_CHANGE`: Of what data type is the `Data` member? `DWORD` as for most other GUIDs or is reading beyond the `UCHAR` from the [`POWERBROADCAST_SETTING`](https://learn.microsoft.com/en-us/windows/win32/api/winuser/ns-winuser-powerbroadcast_setting) struct invalid?


Enyium:

stevewhims:
For all of the assigned writers, please feel free to ignore this. I can process it to completion.

Thanks!
-Steve

stevewhims (approved):

PR #1773, opened 2024-01-02
More info in `power-setting-guids.md`

Enyium:
This is a *possible* correction, after the unresolved questions in #1772 weren't addressed.

- Please verify it for correctness.
- It would also be nice, if `PowerUserNotPresent` could get a description. I can't provide it.

The unresolved questions were:

> - `GUID_GLOBAL_USER_PRESENCE` and `GUID_SESSION_USER_PRESENCE`: What's with the other values from the `USER_ACTIVITY_PRESENCE` enum? Are they also possible values? Probably only `PowerUserNotPresent`.
> - `GUID_LIDSWITCH_STATE_CHANGE`: Of what data type is the `Data` member? `DWORD` as for most other GUIDs or is reading beyond the `UCHAR` from the [`POWERBROADCAST_SETTING`](https://learn.microsoft.com/en-us/windows/win32/api/winuser/ns-winuser-powerbroadcast_setting) struct invalid?

EDIT: We also need a clarification I can't provide: With `GUID_CONSOLE_DISPLAY_STATE` (and probably also `GUID_SESSION_DISPLAY_STATUS`), I get an initial event that informs me about the current monitor state, although the monitor state didn't change. *Is that initial change-unrelated message guaranteed to be sent* (so one can reliably ignore the first message)?

### DIFF
--- a/desktop-src/Power/power-setting-guids.md
+++ b/desktop-src/Power/power-setting-guids.md
@@ -94,6 +94,8 @@ The **Data** member has no information and can be ignored.
 
 The state of the lid has changed (open vs. closed). The callback won't be called until a lid device is found and its current state is known.
 
+The **Data** member is a **DWORD** that indicates the current lid state:
+
 0x0 - The lid is closed.
 
 0x1 - The lid is opened.
--- a/desktop-src/Power/power-setting-guids.md
+++ b/desktop-src/Power/power-setting-guids.md
@@ -75,9 +75,9 @@ The **Data** member is a **DWORD** with a value from the **USER_ACTIVITY_PRESENC
 
 **PowerUserPresent** (0) - The user is present in any local or remote session on the system.
 
-**PowerUserNotPresent** (1)
+**PowerUserNotPresent** (1) - The user is not present in any local or remote session on the system.
 
-**PowerUserInactive** (2) - The user is not present in any local or remote session on the system.
+**PowerUserInactive** (2) - The user is not active in any local or remote session on the system.
 
 ---
 
--- a/desktop-src/Power/power-setting-guids.md
+++ b/desktop-src/Power/power-setting-guids.md
@@ -68,7 +68,7 @@ This notification is sent only services and other programs running in session 0.
 
 **Windows 7, Windows Server 2008 R2, Windows Vista and Windows Server 2008:** This notification is available starting with Windows 8 and Windows Server 2012.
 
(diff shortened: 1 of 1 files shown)


Karl-Bridge-Microsoft (approved):

Enyium:
For the record, these requests again weren't addressed, but ignored:

- The request regarding `GUID_SESSION_USER_PRESENCE` to document `PowerUserNotPresent`. (Only `GUID_GLOBAL_USER_PRESENCE` was addressed.)
- "`GUID_LIDSWITCH_STATE_CHANGE`: Of what data type is the `Data` member? `DWORD` as for most other GUIDs or is reading beyond the `UCHAR` from the [`POWERBROADCAST_SETTING`](https://learn.microsoft.com/en-us/windows/win32/api/winuser/ns-winuser-powerbroadcast_setting) struct invalid?"
  - As a workaround, not knowing the data type, and because Windows uses little endian and the possible values are small, you could just always limit your reads to the `UCHAR`.
- "We also need a clarification I can't provide: With `GUID_CONSOLE_DISPLAY_STATE` (and probably also `GUID_SESSION_DISPLAY_STATUS`), I get an initial event that informs me about the current monitor state, although the monitor state didn't change. *Is that initial change-unrelated message guaranteed to be sent* (so one can reliably ignore the first message)?"

Karl-Bridge-Microsoft:
As these are questions/comments, they need to be included in an Issue or we need to create a private PR to address. I can only review and merge the content revisions you submit in the PR. I've noted your questions and am tracking them internally. Thanks!

Enyium:
For reference:

@Enyium:

> "We also need a clarification I can't provide: With `GUID_CONSOLE_DISPLAY_STATE` (and probably also `GUID_SESSION_DISPLAY_STATUS`), I get an initial event that informs me about the current monitor state, although the monitor state didn't change. _Is that initial change-unrelated message guaranteed to be sent_ (so one can reliably ignore the first message)?"

The remarks for the windowless [`PowerSettingRegisterNotification()`](https://learn.microsoft.com/en-us/windows/win32/api/powersetting/nf-powersetting-powersettingregisternotification#remarks) read (excerpt):

> Immediately after registration, the callback will be invoked with the current value of the power setting.

This makes it stand to reason that this also applies to [`RegisterPowerSettingNotification(…, DEVICE_NOTIFY_WINDOW_HANDLE)`](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-registerpowersettingnotification), which I used when creating this PR.

PR #1779, opened 2024-01-15
Update dx-graphics-hlsl-packing-rules.md

mnicolella:
I don't understand the example as given and believe it is written in error. The documentation earlier in this page states 'every element in an array is stored in a four-component vector.' So why is the problematic example stated as float4[16] when that already is a completely packed array of 64 floats? Also, the modification isn't clear - why is it declaring a 'static' array and trying to cast the original array? float2[32] contains enough space for 128 floats due to each float2 getting padded out to the size of a float4. I think it is clearer to simply offer a rewritten definition of the array.

### DIFF
--- a/desktop-src/direct3dhlsl/dx-graphics-hlsl-packing-rules.md
+++ b/desktop-src/direct3dhlsl/dx-graphics-hlsl-packing-rules.md
@@ -131,7 +131,7 @@ You could pack an array more aggressively. For instance, given an array of float
 
 
 ```
-float4 array[16];
+float array[64];
 ```
 
 
@@ -140,7 +140,7 @@ You could choose to pack it like this, without any spaces in the array:
 
 
 ```
-static float2 aggressivePackArray[32] = (float2[32])array;  
+float4 aggressivelyPackedArray[16];  
 ```
 
 


stevewhims:
Thanks, @mnicolella, I'm investigating.

tex3d:
> I don't understand the example as given and believe it is written in error. The documentation earlier in this page states 'every element in an array is stored in a four-component vector.' So why is the problematic example stated as float4[16] when that already is a completely packed array of 64 floats? Also, the modification isn't clear - why is it declaring a 'static' array and trying to cast the original array? float2[32] contains enough space for 128 floats due to each float2 getting padded out to the size of a float4. I think it is clearer to simply offer a rewritten definition of the array.

I think you misinterpreted the intent.

The problem isn't declaring a `float4[16]` in your constant buffer.  This is part of the demonstrated technique of working around the array element padding.

I agree the document isn't very clear, since it doesn't show a before and after for the element padding workaround.  Let me attempt to demonstrate what it could have said:

Imagine you want to access an array like this from the constant buffer:
```
// Original array, not efficiently packed.
float2 myArray[32];
```

In the constant buffer, this will consume 32 4-element vectors because each array element will be placed at the beginning of one of these vectors.  If you want these values packed tightly in the constant buffer, and still want to access the array in the shader as a float2[32] array, you could write this instead:

```
float4 packedArrayInCBuffer[16];
// shader uses myArray here:
static const float2 myArray[32] = (float2[32])packedArrayInCBuffer;
```

The tighter packing is a trade off versus the need for additional shader instructions for address computation.

Does this make more sense?

stevewhims:
Thanks, all. I've added the info from @tex3d to the topic now. :)

-Steve

PR #1788, opened 2024-01-26
Fixing [Linking Implicitly with a DLL] broken link

lorenzored98:


stevewhims (approved):

stevewhims:
Hi, it looks like the CLA hasn't been signed yet (please see comment in the thread above). If you could sign that, then that'd be great for future PRs. In this case, since I made a lot of changes to the PR myself, and the only change of yours is to change a url, which I've verified to be correct, I'll go ahead and merge this. But if you can sign the CLA for future, that'd be awesome.

Thanks! :)

PR #1797, opened 2024-02-09
Update changing-a-groupampaposs-scope-or-type.md

HerbertMauerer:
couple of clarifications:
- only DLGs may contain FSPs as members.
- Scope of Sid of DLGs
- Conversion hint for global<->domain local group conversions


alvinashcraft (approved):
Thanks for the updates! This looks good.

PR #1802, opened 2024-02-14
Update constants for CryptEncodeObject and CryptDecodeObject

Endecs:
Added missing parameter TIMESTAMP_INFO and its returning structure CRYPT_TIMESTAMP_INFO.


alvinashcraft (approved):
Thanks for adding this item. Looks good.

PR #1809, opened 2024-02-29
Creating contextual TOC for the recent What's New for Windows Server change

Heidilohr:
- Updating breadcrumb file with contextual TOC node to keep What's New in Windows Server 2022 link in correct TOC context.

### DIFF
--- a/desktop-src/breadcrumb/toc.yml
+++ b/desktop-src/breadcrumb/toc.yml
@@ -717,6 +717,9 @@
                     - name: Windows Deployment Services
                       tocHref: /windows/win32/wds/
                       topicHref: /windows/win32/wds/windows-deployment-services-portal
+                    - name: What's New
+                      tocHref: /windows-server/get-started/
+                      topicHref: /windows-server/get-started/whats-new-in-windows-server-2022
                     - name: Windows Management Instrumentation
                       tocHref: /windows/win32/wmisdk/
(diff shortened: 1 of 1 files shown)


Heidilohr:
@GrantMeStrength I'm trying to make sure the link to What's New in Windows Server 2022 keeps customers within the context of the original win32 Windows Server TOC and avoids confusion. Please let me know if the code changes I made to the breadcrumb file are in the wrong place or aren't working as intended.

Heidilohr:
@GrantMeStrength Have you had a chance to take a look at this pull request yet? Please let me know if I got the code incorrect.

Heidilohr:
@stevewhims Please take a look at this pull request when you have time. Thank you!

stevewhims:
Seems we should take this, even though I don't know what a "contextual" TOC is (the one in the "breadcrumb" folder, as opposed to the TOC file that's in the root). But I would have thought that the one in the root would need to be updated.

The guidelines call for sentence case in titles, so this really should be "What's new". But then so many titles around the place don't follow that rule.

Anyway I always feel it's best to take changes quickly; folks can always do fast follow-up fixes if they're necessary.

Also note that this repo is for members of the public to contribute via. For Microsoft folks, we prefer them to use the private repo. In this case that's https://github.com/MicrosoftDocs/win32-pr.

stevewhims (approved):

PR #1811, opened 2024-03-03
Explain debug privilege better

henke37:
This makes it clearer how the privilege works in practice.


alvinashcraft (approved):
Thanks for this update. This looks good.

PR #1814, opened 2024-03-08
Update using-dsaddsidhistory.md

HerbertMauerer:
be extra specific about source domain user role.


alvinashcraft (approved):
Thank you for improving the docs. This looks good.

PR #1816, opened 2024-03-11
Update handling-video-device-loss.md

iPhilip:
Fixed example code so that it runs correctly.


drewbatgit:
@iPhilip Sorry for the delayed response here. I ran your changes by the feature team and they have asked me to reject your proposed changes for these reasons:

1.	Changing the category could cause user to miss dshow filter sources representing a camera like OBS
2.	Only handling DBT_DEVICEREMOVECOMPLETE might be too strict, DBT_DEVICEREMOVEPENDING could put part of the driver in a nonfunctional state.
3.	As for notification, all the “remove” messages should be caught and treated as “this device is not going to be in a working state” message.

PR #1822, opened 2024-03-22
Update r-private-information.md

CryingWelkin:
Added a brief description to the property set, similar to the one of [Public-Information](https://learn.microsoft.com/en-us/windows/win32/adschema/r-public-information).

### DIFF
--- a/desktop-src/ADSchema/r-private-information.md
+++ b/desktop-src/ADSchema/r-private-information.md
@@ -17,7 +17,7 @@ ms.date: 05/31/2018
 
 # Private-Information property set
 
-
+Property set containing user attributes that describe user private information.
 
 | Entry | Value |
 |--------------|--------------------------------------|


CryingWelkin:
[Terminal-Server-License-Server](https://learn.microsoft.com/en-us/windows/win32/adschema/r-terminal-server-license-server) and [MS-TS-GatewayAccess](https://learn.microsoft.com/en-us/windows/win32/adschema/r-ms-ts-gatewayaccess) property sets also lack a brief description. Let me know if that is intended, if not, I can add to them a brief explanation.

alvinashcraft (approved):
This looks good. Thank you for adding the description. I don't think omitting a description was intentional. If you would like to add them for the other two pages in another PR, it would be greatly appreciated. Thanks!

CryingWelkin:
@alvinashcraft Got it. BTW, just wanted to thank you (and your team) for the great effort put into these documnetations. Everything is so vivid in them, such that it is difficult _not to_ understand something. Probably the best documnetation for a technical product, ever.

PR #1827, opened 2024-04-03
Fix table formatting in desktop-src/medfound/direct3d-12-video-structures.md

W0ni:
The html comment breaks the syntax of the markdown table.
This is fixed by removing the newline before the comment.


drewbatgit:
@W0ni Thanks for catching this. I went ahead and fixed all of the formatting issues in this table in a separate PR, so I'm going to close this one. Also, the missing API has been removed from the header, so I went ahead and removed it from the table all together.  Thanks again for helping make the docs better!

PR #1834, opened 2024-04-24
Update a-accountexpires.md

amaiman:
Add additional explanation for the possible values when an account does not expire.


alvinashcraft (approved):
Looks good. Thanks for adding this context.

PR #1835, opened 2024-04-28
OpenGL: Specify `wglCopyContext()` as an actual counterpart of `glXCopyContext()`

cher-nov:
Seems like a documentation mistake to me. Also fixes a lot of excessive spaces in markup.


cher-nov:

stevewhims (approved):

PR #1839, opened 2024-05-01
Update WinHTTP option-flags.md with TLS version configuration

ericlaw1979:
Correct discussion of the WinHTTP_OPTION_SECURE_PROTOCOLS option. The old explanation wasn't correct.


stevewhims (approved):

stevewhims:
PS we encourage internal Microsoft folks to contribute to the private content repos, rather than the public ones like this. :)

PR #1840, opened 2024-05-02
Fix debugging error message

MajorMurphy:
Printing correct name of failed function


alvinashcraft (approved):
Thanks for catching this. Looks good!

PR #1844, opened 2024-05-16
Update processor-groups.md

Rblume-Microsoft:
This update was made after the servicing Bug 48960313


Karl-Bridge-Microsoft:
Closing as this is not a dev doc issue. 

As it seems to be a system/platform bug with an upcoming fix, I suggested a KnowledgeBase article. Email with details sent to OP.

PR #1846, opened 2024-05-18
pe-format: MIPS Updates

FlyGoat:
Hi,

This PR updated pe-format `Machine Types` for MIPS using known values from https://learn.microsoft.com/en-us/windows/win32/sysinfo/image-file-machine-constants and added clarification text for those values.

We further request Microsoft assistance on two matters:

1. **Allocation of new values**: We would like Microsoft to allocate Machine Type following values to cover more existing architectures:
- `IMAGE_FILE_MACHINE_R4000BE`: MIPS III compatible 64-bit big endian, our educated guess of allocation to that value is `0x164`, this is necessary to cover 64-bit big endian MIPS systems.
- `IMAGE_FILE_MACHINE_MIPS32R6BE`, `IMAGE_FILE_MACHINE_MIPS32R6LE`, `IMAGE_FILE_MACHINE_MIPS64R6BE`, `IMAGE_FILE_MACHINE_MIPS64R6LE`: MIPS Release 6 32/64 bit big/little endian. MIPS Release 6 is no longer backwards compatible with previous releases of the architecture, thus it is necessary to mark them with new machine type. No new relocation type is required.

2. **Clarification on header endianness**: From online sources and observations made on existing Big Endian PE images, we learned that in Microsoft's implementation, headers in PE image are always stored as little-endian byte order regardless of targeting architecture endianness. We would like Microsoft to confirm this behaviour and document it in pe-format if possible.

This PR has similar nature to #1017, please handle it accordingly.

Thanks


FlyGoat:
ping @Karl-Bridge-Microsoft

Karl-Bridge-Microsoft (approved):

PR #1848, opened 2024-05-21
Fix YAML parsing errors

W0ni:
Some strings in the YAML part of the markdown files are invalid, because of the usage of single quotes, which should be replaced by double quotes.

For instance, single quotes were used inside single quotes. Moreover, double quotes allow the `\_` to be "parsed".


GrantMeStrength:
Thanks for the fixes!

PR #1856, opened 2024-05-30
Add lifetime info for system surrogate

brialmsft:
I suspect there's a more maintainable way to do the AgileReference link, but it seems like simple label-based linking doesn't work because the target page is in a different subdirectory.


brialmsft:
@hart-wilson does this seem like a reasonable level of detail to document for dllhost lifetime?

dmachaj (approved):

stevewhims (approved):

PR #1858, opened 2024-06-02
Fix wrong word in Subject Unique identifier ASN.1

NoahFreising:
Fixes a small copy and paste error.


alvinashcraft (approved):
Thanks for catching this issue. Looks good.

PR #1866, opened 2024-06-14
Fix a number of misspellings

Blake-Madden:
Note that there were a couple of `_tcsat` that I changed to `_tcscat` (the correct function name).


GrantMeStrength:
Thank you so much! Lots of fixes! -John

PR #1872, opened 2024-07-03
Update add-tree-view-items.md

gtworek:
The sample is wrong. Sizeof(LPWSTR) returns size of a pointer, and not length of the string.

### DIFF
--- a/desktop-src/Controls/add-tree-view-items.md
+++ b/desktop-src/Controls/add-tree-view-items.md
@@ -55,7 +55,7 @@ HTREEITEM AddItemToTree(HWND hwndTV, LPTSTR lpszItem, int nLevel)
 
     // Set the text of the item. 
     tvi.pszText = lpszItem; 
-    tvi.cchTextMax = sizeof(tvi.pszText)/sizeof(tvi.pszText[0]); 
+    tvi.cchTextMax = _tcslen(tvi.pszText) + 1; 
 
     // Assume the item is not a parent item, so give it a 
     // document image. 


GrantMeStrength (changes requested):
The change from `sizeof(tvi.pszText)/sizeof(tvi.pszText[0])` to `_tcslen(tvi.pszText) + 1` both have issues. The original is wrong because `pszText` is a pointer, so `sizeof` gives the pointer size (4 or 8), not the buffer size. The replacement uses the string length as the buffer size, but `cchTextMax` should reflect the *allocated buffer capacity*, not the current string length — otherwise any text the user types that's longer than the initial string will be silently truncated. Please use the actual buffer allocation size, e.g. `tvi.cchTextMax = MAX_PATH;` if that matches your buffer, or document the buffer size explicitly.

GrantMeStrength:
Thanks for catching this, @gtworek — you're absolutely right that `sizeof(tvi.pszText)/sizeof(tvi.pszText[0])` is wrong because `pszText` is a pointer, not an array.

We're going to fix this internally with a slightly different approach: for `TVM_INSERTITEM` (and `TreeView_SetItem`), `cchTextMax` is actually ignored by the control — it's only used when *getting* item text. So we'll update the sample to clarify that, rather than just swapping in `_tcslen`.

Closing this PR since we'll apply the fix through our internal workflow. Appreciate the contribution!

GrantMeStrength:
@copilot /review

GrantMeStrength (approved):
**Review: Approved ✅**

This is a legitimate bug fix. The original code `sizeof(tvi.pszText)/sizeof(tvi.pszText[0])` computes the size of a *pointer* divided by the size of a TCHAR — yielding 4 or 8, not the actual string length. The fix `_tcslen(tvi.pszText) + 1` correctly computes the string length including the null terminator.

Good catch.

PR #1877, opened 2024-07-10
[HLSL] Intrinsic doc update

farzonl:


desktop-src/direct3dhlsl/AddUint64.md
```diff
@@ -0,0 +1,61 @@
...
+
+```syntax
+uint<c> AddUint64(uint<c> a, uint<c> b);
+```
+
+## Parameters
+
+| Item | Description |
+|------|-------------|
+| *a* | [in] A vector of type uint64_t (64-bit unsigned integer) as the first operand.  || *b* | [in] A second vector of type uint64_t (64-bit unsigned integer) as the second operand.  |
```
bogner: This table (and many of the later ones) seems to have lost some new lines between rows and doesn't render correctly.

desktop-src/direct3dhlsl/AddUint64.md
```diff
@@ -0,0 +1,61 @@
...
+## Return value
+
+ Returns a vector resulting from the addition of the two input vectors. Each component in the resulting vector is the sum of the corresponding components of the input vectors. All operations are performed on uint64_t types.
+## Type Description
+
+| Name  | [**Template Type**](../direct3dhlsl/dx-graphics-hlsl-data-types.md)| [**Component Type**](../direct3dhlsl/dx-graphics-hlsl-data-types.md) | Size |
+|-------|--------------------------------------------------------------------|----------------------------------------------------------------------|------|
+| *ret*   | [**vector**](../direct3dhlsl/dx-graphics-hlsl-vector.md) | [**uint**](../WinProg/windows-data-types) | any |
+| *a*   | [**vector**](../direct3dhlsl/dx-graphics-hlsl-vector.md) | [**uint**](../WinProg/windows-data-types) | any |
+| *b*   | [**vector**](../direct3dhlsl/dx-graphics-hlsl-vector.md) | [**uint**](../WinProg/windows-data-types) | any |
```
bogner: Size should be 2 or 4 here (See "AddUint64 can only be applied to uint2 and uint4 operands" error in HLOperationLower.cpp)

desktop-src/direct3dhlsl/AddUint64.md
bogner: These aren't vectors of type uint64_t, they're vectors of uint and we're adding them two-by-two.

desktop-src/direct3dhlsl/AddUint64.md
```diff
@@ -0,0 +1,61 @@
...
+api_name:
+- AddUint64
+api_type:
+- NA
+---
+
+
+# AddUint64
+
+Unsigned add of 32-bit operand with the carry.
```
bogner, line 19: This description is pretty unclear. Probably just have to write one up manually.

desktop-src/direct3dhlsl/QuadAny.md
```diff
@@ -0,0 +1,66 @@
...
+```
+
+## Parameters
+
+| Item | Description |
+|------|-------------|
+| *cond* | [in] A vector or scalar variable of type boolean which we want to evaluate. QuadAny will return true if any component of this variable is true.  |
+## Return value
+
+ Returns a scalar of type boolean, true if any component of the parameter is true, otherwise it returns false.
```
bogner, line 38: Interesting that it recognized this is about the quad for `all` but not for `any`

desktop-src/direct3dhlsl/WaveMatch.md
```diff
@@ -0,0 +1,76 @@
...
+* [**Hull Shader**](https://learn.microsoft.com/en-us/windows/uwp/graphics-concepts/hull-shader-stage--hs-)
+* [**Domain Shader**](https://learn.microsoft.com/en-us/windows/uwp/graphics-concepts/domain-shader-stage--ds-)
+* [**Geometry Shader**](https://learn.microsoft.com/en-us/windows/uwp/graphics-concepts/geometry-shader-stage--gs-)
+* [**Raygeneration Shader**](../direct3d12/ray-generation-shader.md)
+* [**Intersection Shader**](../direct3d12/intersection-shader.md)
+* [**Anyhit Shader**](../direct3d12/any-hit-shader.md)
+* [**Closesthit Shader**](../direct3d12/closest-hit-shader.md)
+* [**Miss Shader**](../direct3d12/miss-shader.md)
+* [**Callable Shader**](../direct3d12/callable-shader.md)
+* **Node Shader**
```
bogner, line 70: This lists every single shader stage, but other functions that should be valid for all have this section blank. I suspect that means the ones that are blank are ones where the AI just doesn't know the answer.

desktop-src/direct3dhlsl/dot4add_u8packed.md
```diff
@@ -0,0 +1,62 @@
...
+```
+
+## Parameters
+
+| Item | Description |
+|------|-------------|
+| *a* | [in] The first input parameter of scalar and component type uint.  || *b* | [in] The second input parameter of scalar and component type uint.  || *c* | [in] The third input parameter of scalar and component type uint.  |
+## Return value
+
+ Returns a scalar with component type uint. The result is the sum of the products of the corresponding entries of the two input vectors along with the addition of the third parameter.
```
bogner, line 40: Computes dot product, not sum

bogner (commented):
Left some notes on the oddities I noticed skimming through. Ironically the very first one alphabetically is probably the one that's the least correct. Otherwise I only spotted a few places that are glaringly wrong.

desktop-src/direct3dhlsl/AddUint64.md
farzonl: Its rendering correctly in md, because the right number of `|`  but this is a simple fix. in the generator.

desktop-src/direct3dhlsl/AddUint64.md
farzonl: we don't encode that in `DirectXShaderCompiler.utils.hct.hctdb_instrhelp` so my view is just that it is a `uint<c>` which mean this is something that would have to be a manual edit since `hctdb_instrhelp` isn't giving enough detail.

desktop-src/direct3dhlsl/AddUint64.md
farzonl: This is bad AI generation. looks like it is making assumptions.

desktop-src/direct3dhlsl/WaveMatch.md
farzonl: This isn't an llm failure. Some intrinsics have documentation about which shader models are supported and which aren't that is because DXILOps 6-71, 78-80, 101, 130-131, 160, 162-164,  216-221, and 244 never had shader model documentation while the rest do.

PR #1878, opened 2024-07-15
Add HLSL vector/matrix template declaration option

### LINKED ISSUE #2034: generic vector and matrix underspecified
This is not necessarily a bug. Just the doc doesn't say why these expressions are allowed, so we don't know what they do:

`vector v;`
that stuff builds on both DXC and FXC. Please write clearly in the doc what this is.

`matrix m;`
same.

`vector<half> v;`
This builds in DXC, but it was not building in FXC. Clear spec required.
According to:
https://docs.microsoft.com/en-us/windows/desktop/direct3dhlsl/dx-graphics-hlsl-vector
This is not supposed to be legal.

`matrix<half> m;`
Same, refused by FXC but ok for DXC. What is this ? according to tests it appears to be a full 4x4 matrix.

`matrix<half, 3> m;`
This appears to be a matrix of 3 rows and 4 columns. why ? why not square ?
FXC refuses this.

bests.

pow2clk:
Refactors vector and matrix docs pages to allow more self-referencing of concepts, clear up some confusing parts, adds examples, and finally adds the default values of the template parameters that allows them to be excluded to get the defaults, which was the actual motivator of this change.

Fixes https://github.com/microsoft/DirectXShaderCompiler/issues/2034

### DIFF
--- a/desktop-src/direct3dhlsl/dx-graphics-hlsl-matrix.md
+++ b/desktop-src/direct3dhlsl/dx-graphics-hlsl-matrix.md
@@ -17,83 +17,62 @@ api_location:
 
 # Matrix Type
 
-A matrix is a special data type that contains between one and sixteen components. Every component of a matrix must be of the same type.
+A matrix is a data type that contains between one and sixteen [scalar](dx-graphics-hlsl-scalar.md) components in a two dimensional grid. Every component of a matrix must be of the same type.
 
+## Type Declaration
 
+You can declare matrix variables by using the [scalar type](dx-graphics-hlsl-scalar.md) name of the matrix's contents with the number of rows and columns it contains:
 
-|                         |
-|-------------------------|
-| **TypeComponents Name** |
-
-
-
- 
-
-## Components
-
-
-
-| Item                                                                                                                             | Description                                                                                                                                                                                                                                                                                       |
-|----------------------------------------------------------------------------------------------------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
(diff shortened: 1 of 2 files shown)


desktop-src/direct3dhlsl/dx-graphics-hlsl-vector.md
```diff
@@ -16,80 +16,47 @@ ms.date: 05/31/2018
...
- 
-
-## Examples
-
-Here are some examples:
-
-
-```
+```hlsl
 bool    bVector;   // scalar containing 1 Boolean
```
python3kgae: Should this be bool1 instead of bool?

pow2clk: It's pre-existing. I think the intent was to demonstrate the difference between declaring something as a scalar versus as a vector. I don't think that's terribly useful myself, but I'm ambivalent.

pow2clk: I've revised the names and comments

stevewhims (approved):

desktop-src/direct3dhlsl/dx-graphics-hlsl-matrix.md
```diff
@@ -17,83 +17,62 @@ api_location:
...
                                };
+matrix<int16_t, 1>   sMatrix = { 1, 2, 3, 4 }; // Defaults to 1x4 int16 matrix
+matrix<float16_t>    hMatrix = { 0.0f, 0.1f, 0.2f, 0.3f, // Defaults to 4x4 float16 matrix
+                                 1.0f, 1.1f, 1.2f, 1.3f,
+                                 2.0f, 2.1f, 2.2f, 2.3f,
+                                 3.0f, 3.1f, 3.2f, 3.3f }; 
+matrix               fMatrix = { 0.0f, 0.1f, 0.2f, 0.3f, // Defaults to 4x4 float matrix
+                                 1.0f, 1.1f, 1.2f, 1.3f,
+                                 2.0f, 2.1f, 2.2f, 2.3f,
+                                 3.0f, 3.1f, 3.2f, 3.3f };
```
siliconvoodoo, line 73: nice

PR #1882, opened 2024-07-18
Layout fix

doracpphp:
The layout of “Subcategory” under “Auditing Constants” was broken and has been corrected.

## Before
image

URL: https://learn.microsoft.com/en-us/windows/win32/secauthz/auditing-constants

## After

image


alvinashcraft (approved):
Looks good. Thanks for this update to the layout.

PR #1890, opened 2024-08-06
Remove incorrect Buffer GetDimensions exception

pow2clk:
Per https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/sm5-object-buffer-getdimensions Buffers do have this method. This removes the inaccurate exclusion of them from one place the method is documented.


damyanp:
Is this the right place to make this change?  The page title is "GetDimensions (DirectX HLSL Texture Object), and it's a part of the SM4 documentation for Texture Object.

Buffer wasn't added until SM5, I think, and has documentation here: https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/sm5-object-buffer-getdimensions

stevewhims (approved):

pow2clk:
There is no other place that implies that Buffer doesn't have a getDimimensions method. It's weird this page mentioned it in the first place

PR #1894, opened 2024-08-12
Expand on how RTF is streamed out

Blake-Madden:
Clarify that `SF_UNICODE` is not meant for RTF streaming, and explain that standard RTF is 7-bit ASCII.


GrantMeStrength:
Thanks for the excellent clarification, Blake! Recreated in the internal repo for merge.

PR #1897, opened 2024-08-15
Fix wrong field

ghost:


GrantMeStrength:
Researching.
https://stackoverflow.com/questions/28068630/vs-versioninfo-structure-unnecessary-padding

jwmsft (approved):

PR #1904, opened 2024-09-02
GL_EXP2 fog application equation

callumames99:
The square removes the necessary sign in the exponentiation function which causes f to tend towards infinity and not zero.
I have moved the sign to the appropriate position such that it is unaffected by the square and f tends towards zero.


stevewhims:
The original formula is wrong, for sure. But I'm not sure density should be squared. Some sources have e to the (minus density times z-squared). Let me check with the product team.

stevewhims:
Well, we found a source that agrees with your new image, so let's take that. :)

stevewhims (approved):

callumames99:
Thank you for this. I apologise for being unable to locate the original
typefont used to render the formulae prior to this update (and the odds of
finding it are probably slim given these date back to at least 2005).

Kindest Regards.

PR #1907, opened 2024-09-07
Update working-with-strings.md

PeterCon:
The description of UTF-16 fails to distinguish between Unicode characters and UTF-16 code units. This oversight frequently leads to code bugs when handling text because developers end up treating individual surrogate code units as a complete character. Such bugs violate global readiness and internationalization best-practice guidelines.

This change adds a note explaining this distinction and points to another topic with more details.


GrantMeStrength:
Thanks PeterCon! Recreated in the internal repo for merge (with the link path corrected).

PR #1930, opened 2024-10-22
Update changing-the-password-on-a-serviceampaposs-user-account.md

james-msft:
1. correct the input parameter for _tcscat_s
2. need to init com before calling ADsGetObject.


alvinashcraft (approved):
Thanks for fixing the code example here.

PR #1931, opened 2024-10-23
Update prioritizing-schannel-cipher-suites.md

james-msft:
add #pragma comment(lib, "Bcrypt.lib") to enable uses of BCrypt* functions


alvinashcraft (approved):
Looks good. Thanks for the update!

PR #1934, opened 2024-10-31
Update retrieving-a-certificate-revocation-list.md

snakex2000:
Updated error handling for SysAllocString failure case

Removed errant ; in the call to CoCreateInstance


alvinashcraft (approved):
Thanks for catching that semicolon and updating to use a null check. Looks good.

PR #1944, opened 2024-11-13
Correct numeric value of SE_SACL_DEFAULTED

RavuAlHemio:


alvinashcraft (approved):
Good catch. The DACL value was repeated here. I just double checked the values in winnt.h, and your update is correct. Thanks for your contribution!

PR #1951, opened 2024-12-03
Fix incorrect position of FILE_SHARE_WRITE

dskr999:
Fix incorrect position of FILE_SHARE_WRITE for the columns

GENERIC_READ, FILE_SHARE_READ, FILE_SHARE_WRITE

and

GENERIC_WRITE, FILE_SHARE_READ, FILE_SHARE_WRITE

Otherwise it creates a very wrong impression of how sharing flags work


alvinashcraft (approved):
Thank you for this fix. It looks like there was a problem when this table was part of a mass conversion from HTML tables several years ago.

PR #1964, opened 2025-01-10
Add error handling in the FormatMessage + Last Error example

nir9:
Problem: Current example does not consider possible failures of the functions that are called before finally displaying the error message in a box.

Solution: Add error handling to the example

### DIFF
--- a/desktop-src/Debug/retrieving-the-last-error-code.md
+++ b/desktop-src/Debug/retrieving-the-last-error-code.md
@@ -8,21 +8,19 @@ ms.date: 05/31/2018
 
 # Retrieving the Last-Error Code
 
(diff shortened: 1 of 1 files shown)


nir9:
BTW, I think it may be a better idea to simplify this example to the very core idea of just using FormatMessage and printing it out on the screen (without the logic of the string formatting). Let me know what you think of this idea, and if yes, I can amend this.

Karl-Bridge-Microsoft:
@nir9 - simplifying the example is a good idea.

nir9:
@Karl-Bridge-Microsoft done

Karl-Bridge-Microsoft (approved):

RushilJalal:
@nir9 Hey Nir, I wanted to ask something about debugging C socket programming code using gdb. I am unable to step into functions like `socket()`. I am presented with an error 'Could not load source './socket/../sysdeps/unix/syscall-template.S': 'SourceRequest' not supported.' I am sure this is not the right place to ask this but I'm not aware of a better one. Thank you.

PR #1979, opened 2025-02-04
overviewTypoFix

cchavez-msft:


alvinashcraft (approved):
Thanks for the fix. Looks good.

PR #1983, opened 2025-02-14
Update ray flags documentation

mmoult:
Include a description for the final two ray flag enumerations, SKIP_TRIANGLES and SKIP_PROCEDURAL_PRIMITIVES. The description for each was sourced from the DirectX specs, specifically:
https://microsoft.github.io/DirectX-Specs/d3d/Raytracing.html


mmoult:

stevewhims (approved):

PR #1986, opened 2025-02-23
Resolve confusing code, total of 20 writes or 20 writes per thread

AnishKacham:
The code example is slightly confusing. 
Currently, each thread writes to the DB 20 times, i.e, a total 40 writes to the DB.
I think the example is supposed to illustrate how two threads together can do 20 writes to a DB (interleavingly)

### DIFF
--- a/desktop-src/Sync/using-mutex-objects.md
+++ b/desktop-src/Sync/using-mutex-objects.md
@@ -26,7 +26,7 @@ If a mutex is abandoned, the thread that owned the mutex did not properly releas
 #define THREADCOUNT 2
 
 HANDLE ghMutex; 
-
+DWORD dwCount = 0;
 DWORD WINAPI WriteToDatabase( LPVOID );
 
 int main( void )
@@ -86,7 +86,7 @@ DWORD WINAPI WriteToDatabase( LPVOID lpParam )
     // lpParam not used in this example
     UNREFERENCED_PARAMETER(lpParam);
 
-    DWORD dwCount=0, dwWaitResult; 
+    DWORD dwWaitResult; 
 
     // Request ownership of mutex.
 


stevewhims:
Thanks, @AnishKacham! I'm not seeing any evidence that the sample has a goal of 20 writes total. This change achieves that hypothetical goal, but only for the special case of 2 threads. So it seems like an arbitrary change. Wouldn't it be more logical to believe that the sample wants 20 writers total no matter how many threads there are? If we leave the code as-is, then no matter what we set THREADCOUNT to, each thread will make 20 writes. So, yes, the total number of writes will be THREADCOUNT * 20. I'm not sure what's wrong with that design. I'll go ahead and close this PR as it doesn't seem to be something we want or need to change. But please let me know if you have more info about why this would be a logical change.

Thanks!
-Steve

AnishKacham:
Hey @stevewhims ,
Thank you for thr review.
The first few lines of the doc talks about accessing shared resource from multiple threads.
> You can use a mutex object to protect a shared resource from simultaneous access by multiple threads or processes. Each thread must wait for ownership of the mutex before it can execute the code that accesses the shared resource.

So when I read the example, I thought it's simulating something like a single database cursor or something which when used by multiple threads simultaneously may cause issues. 
So when I came to the example, I was looking for what was the shared resource that was being protected. And with the current example, because there is no sharing of any resource ('dwCount' is local to the thread) it wouldn't matter if there was no mutex in the first place right?

But I agree with,
> I'm not seeing any evidence that the sample has a goal of 20 writes total

> This change achieves that hypothetical goal, but only for the special case of 2 threads

What I essentially was trying to achieve with this change was to show a shared resource being protected by the mutex. So my goal was not exactly to achieve 20 writes total.

In hindsight I should've explained this better in the description before hand but,
Let me know what you think?

Thanks,
Anish.

stevewhims:
Yes, I see, thanks.

In "You can use a mutex object to protect a shared resource from simultaneous access by multiple threads", I think the operative word is "can". If you have the discipline to know *all* the places in your code where you access that resource, and you guard them all with a mutex, then yes you can do that. If a shared resource were a compound object with the resource itself plus its own built-in mutex, then that would make the design even safer (and the wording would change to "You use this object's mutex to project this same object's shared resource"; then the word "can" would go away). Strictly speaking, what the mutex does is guarantee that only at most one thread at a time is in a section of code. But yes, a practical *use* to put that guarantee to is "to protect a shared resource". But folks find all kinds of uses for all kinds of useful things.

And then in "For example, if several threads share access to a database, the threads can use a mutex object to permit only one thread at a time to write to the database", I think the operative word is "example". It's just an example. The shared resource in that case is the database itself (whether that's a good example or not is another matter). But *that* is what the shared resource is. The storage containing `dwCount` is not the shared resource, AFAIK. Otherwise I think the topic would have said, "For example, if several threads share access to a variable, the threads can use a mutex object to permit only one thread at a time to access the variable."

So sure enough the mutex really does make sure that only one thread at a time can execute the `printf` in the code listing. And it's that `printf`, I think, that represents a write to the database (as the comment above it suggests). I think it's a red herring that the count is incremented inside the protection afforded by the mutex. We use that count only to scale up the writes (to give each thread 20 writes). If the reader had the faith that the hypothetical database write were substantial enough (took a substantial amount of time) and that there were enough threads, and that each thread were hitting the database often enough, then I think we could omit that count in the example. But it's in there, for better or worse.

Anyway, I still think that this topic is ok as-is. It's highlighting a (or *the*, if you prefer) use case for a mutex, but it's not saying that that's *all* they can be used for (someone with enough imagination will always be able to come up with uses for things that we didn't think of, so it's never safe saying "this is the thing it's used for"). And I think it's clear that a database is just an example of a shared resource. Again, it's not the only kind of shared resource. And as I say, that may not be the best example, but I think it makes the point reasonably well. Especially when in reality the mutex in this example is protecting nothing but a printf and a comment! :) So obviously it's illustrating a point, and not a real-world, production-ready example. I think/hope the reader will know when they need to protect a shared resource, and will by emulating this example know how to protect their code that accesses that resource.

PR #1989, opened 2025-02-26
Clean up superfluous semicolons

Rageking8:
Summary:
- Remove extraneous semicolons in code snippets
- Move `title` metadata to the top
- Elide excess trailing empty lines (at the end of the page) which renders as a huge empty space

### DIFF
--- a/desktop-src/directcomp/initialize-directcomposition.md
+++ b/desktop-src/directcomp/initialize-directcomposition.md
@@ -544,7 +544,7 @@ HRESULT DemoApp::Initialize()
     wcex.cbClsExtra    = 0;
     wcex.cbWndExtra    = sizeof(LONG_PTR);
     wcex.hInstance     = HINST_THISCOMPONENT;
-    wcex.hbrBackground = (HBRUSH)(COLOR_WINDOW+1);;
+    wcex.hbrBackground = (HBRUSH)(COLOR_WINDOW+1);
     wcex.lpszMenuName  = nullptr;
     wcex.hCursor       = LoadCursor(NULL, IDC_ARROW);
     wcex.lpszClassName = L&quot;DirectCompDemoApp&quot;;
@@ -981,7 +981,3 @@ HRESULT DemoApp::MyCreateGDIRenderedDCompSurface(HBITMAP hBitmap,
 
 [**SafeRelease**](/windows/desktop/medfound/saferelease)
 </dt> </dl>
-
- 
-
- 
--- a/desktop-src/Controls/create-a-simple-combo-box.md
+++ b/desktop-src/Controls/create-a-simple-combo-box.md
@@ -282,7 +282,7 @@ HRESULT DemoApp::Initialize()
     wcex.cbClsExtra    = 0;
     wcex.cbWndExtra    = sizeof(LONG_PTR);
     wcex.hInstance     = HINST_THISCOMPONENT;
-    wcex.hbrBackground = (HBRUSH)(COLOR_WINDOW+1);;
+    wcex.hbrBackground = (HBRUSH)(COLOR_WINDOW+1);
     wcex.lpszMenuName  = NULL;
     wcex.hCursor       = LoadCursor(NULL, IDC_ARROW);
     wcex.lpszClassName = TEXT("DemoApp");
@@ -492,7 +492,3 @@ LRESULT CALLBACK DemoApp::WndProc(HWND hwnd, UINT message, WPARAM wParam, LPARAM
 
 [ComboBox](combo-boxes.md)
 </dt> </dl>
-
- 
-
- 
--- a/desktop-src/Controls/create-directory-listing-in-a-single-selection-list-box.md
+++ b/desktop-src/Controls/create-directory-listing-in-a-single-selection-list-box.md
@@ -81,7 +81,7 @@ INT_PTR CALLBACK DlgDelFileProc(HWND hDlg, UINT message,
                     iRet = MessageBox(hDlg, achTemp, L"Deleting Files", 
                         MB_YESNO | MB_ICONEXCLAMATION);
                     if (iRet == IDNO)
-                        return TRUE;;
+                        return TRUE;
 
                     // Delete the file.
                     fResult = DeleteFile(pszFileToDelete); 
@@ -146,7 +146,3 @@ INT_PTR CALLBACK DlgDelFileProc(HWND hDlg, UINT message,
 
 [Using List Boxes](using-list-boxes.md)
 </dt> </dl>
-
- 
-
- 
--- a/desktop-src/CoreAudio/render-spatial-sound-using-spatial-audio-objects.md
+++ b/desktop-src/CoreAudio/render-spatial-sound-using-spatial-audio-objects.md
@@ -1,7 +1,7 @@
 ---
+title: Render Spatial Sound Using Spatial Audio Objects
 description: This article presents some simple examples that illustrate how to implement spatial sound using static spatial audio objects, dynamic spatial audio objects, and spatial audio objects that use Microsoft's Head Relative Transfer Function (HRTF).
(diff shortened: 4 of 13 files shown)


stevewhims:
Hi, @Rageking8!

Thank you for the intent here, which is to improve and correct the docs. We appreciate that. These are changes that I think we can do more efficiently in-house, though. We could do them via scripting (or equivalent automation) across entire repos in one shot, rather than piecemeal a few topics at a time. For example, there are likely to be many tens of thousands of topics where the title isn't at the top of the YAML frontmatter. If I wrote a script to remedy that, then once I'd tested my logic I could make those many tens of thousands of deltas in one shot with confidence. Whereas if we do that same work via public contributions like this, then in order to be safe and duly diligent, we'd have to read through every one of those many tens of thousands of deltas; which of course would not be good ROI.

In addition, these specific changes here don't materially affect the value of the content, so it's not entirely clear that we want to make them at all (considering other more clearly high priority tasks we could be spending that same time on).

Let's go ahead and close this PR without taking it.

Thanks again!
-Steve

Rageking8:
Hi @stevewhims, would a PR with just the content changes be desirable (i.e. the removal of extra semicolons and empty lines)? Since I would be keen to learn which changes are accepted and merged (maybe typo fixes)?

stevewhims:
> Hi @stevewhims, would a PR with just the content changes be desirable (i.e. the removal of extra semicolons and empty lines)? Since I would be keen to learn which changes are accepted and merged (maybe typo fixes)?

So, across the repos that I have cloned, and that I work in a lot, I see 1,590 instances of double semicolon. Rather than fix the cases in a single topic, or in many topics over time and over many PRs, I'd prefer to just fix the cases across all of those in one PR per repo. If I were to script fixes to all those, then I think I'd need to test a reasonable cross-section of cases (maybe 10 of them), and then pull the trigger and update all 1,590. I might look at 10 changes in the PR(s), and take the other 1,580 on faith (and get them for free). If a 1,590-change PR comes in from outside of my team then it'd be only good practical sense and hygiene for me to sanitize every one of those changes (in the same way that any secure app will sanitize input from end users). It's not about lack of trust, it's not about expecting malice, it's simply about best practices, and not taking unnecessary chances. So needless to say, doing something like that myself would be far more efficient than having an external contributor do it.

But across the repos that I have cloned, and that I work in less often, I see 884,000 instances of double semicolon! So now the work becomes investigating a good cross-section of those cases to see what's going on, figuring out which ones are wrong, and then deciding whether it's worth fixing them. Right now it's not clear to me that fixing double semicolons is compelling at all, given other higher pri work we have waiting to be done, given that they'll compile just fine.

As for typos, it takes only about 20-30 seconds for a writer to open a PR, check that all the changes are just typo fixes, and that they really are fixing the typo and not making anything worse, and then click the merge button, wait for it to finish, then delete the branch where necessary. That's plus the time the originator took to make the PR in the first place. That investment weighs against the possibility that anyone actually stumbled over the typo in the first place enough for it to make a material difference to their comprehension of the content. Is the typo really wasting customers' time understanding the topic, or is the typo more of an annoyance that somebody can't resist wanting to fix? I know that typos annoy me, and I always feel better after fixing them. But is that really (always) good ROI? I don't think it's clear. So while I like fixing typos when I see them, when I'm already in a topic anyway, when it's super cheap (I can fix a typo sub-second if it's right in front of me), I generally consider external typo-fixing PRs to be more expensive than that (30 seconds or so, and so I worry very slightly about them). But it's not a biggie either way. :)

I think that the value of external PRs is when some content was really clearly materially improved for a large number of readers. I could ask myself for any PR I submit, who am I doing this for, and why? Is it just because that comma or that word or that typo or what whitespace annoyed me, and I couldn't resist the urge to fix it (even though it was easy to read even with the typo, and only one person a year looks at that topic)? Was that really a great use of my time (and the time of everyone else I'm roping in to process my PR)? Or did I make the PR because I geniunely believe that the content I've fixed was previously incorrect, misleading, or significantly unclear enough to most readers (out of a large audience of readers) to be worth the fix? Remember, we have hundreds of thousands of topics to maintain, and most of them are long tail in terms of page views. Important changes/fixes to important topics (or new, important topics) are where the real value is.

Thanks!
-Steve

Rageking8:
Hi @stevewhims, I just have a few points to mention:

- If we take this PR for example (which is mid sized), I would presume it takes less than 5 minutes for a thorough review. Although I do agree the impact is negligible at best, since the extra semicolons and lines are non-critical for the content, and the topics involved might have low viewers. But I do believe in incremental improvements to the docs, and 5 mins is a small enough "investment" that I don't think closing the PR is massively more compelling.

- Given the mention of fixing all occurrences of a particular fault in a repo at once (with or without automation), this PR *does* exhaustively fix all instances, by using the regex `;;$` (searching for ";;" has way too many false positives), and manually reviewing all to find high confidence positive hits.

In conclusion, from my understanding, the only worthwhile external PRs are ones that add valuable content, correct non-factual content, or change content in such a way it's much better for the reader. In addition, the topic needs to be relevant (not outdated or archived), with a sizable viewership as well, to justify the cost of everyone involved. As an aside, these PRs also tend to take longer to review and fact-check.

Hence, @stevewhims, thanks for bringing me up to speed with the concerns behind reviewing and merging certain PRs. I would examine the criteria, should I decide to make another one in the future. Thanks again!

desktop-src/directcomp/initialize-directcomposition.md
```diff
@@ -544,7 +544,7 @@ HRESULT DemoApp::Initialize()
     wcex.cbClsExtra    = 0;
     wcex.cbWndExtra    = sizeof(LONG_PTR);
     wcex.hInstance     = HINST_THISCOMPONENT;
-    wcex.hbrBackground = (HBRUSH)(COLOR_WINDOW+1);;
+    wcex.hbrBackground = (HBRUSH)(COLOR_WINDOW+1);
     wcex.lpszMenuName  = nullptr;
     wcex.hCursor       = LoadCursor(NULL, IDC_ARROW);
     wcex.lpszClassName = L&quot;DirectCompDemoApp&quot;;
```
Rageking8, line 550: I would just throw in a quick last question. Since the line above contains an error that would make the code snippet fail to compile (the `&quot;`), I would assume it's more efficient to do the fix en masse internally.

Hence, am I right that an external PR to fix such issues (regardless of whether said PR fixes all occurrences) is not recommended?

stevewhims: I feel like it'd be more efficient to fix those HTML entity problems internally.

stevewhims (approved):

stevewhims:
> Hi @stevewhims, I just have a few points to mention:
> 
> * If we take this PR for example (which is mid sized), I would presume it takes less than 5 minutes for a thorough review. Although I do agree the impact is negligible at best, since the extra semicolons and lines are non-critical for the content, and the topics involved might have low viewers. But I do believe in incremental improvements to the docs, and 5 mins is a small enough "investment" that I don't think closing the PR is massively more compelling.
> * Given the mention of fixing all occurrences of a particular fault in a repo at once (with or without automation), this PR _does_ exhaustively fix all instances, by using the regex `;;$` (searching for ";;" has way too many false positives), and manually reviewing all to find high confidence positive hits.
> 
> In conclusion, from my understanding, the only worthwhile external PRs are ones that add valuable content, correct non-factual content, or change content in such a way it's much better for the reader. In addition, the topic needs to be relevant (not outdated or archived), with a sizable viewership as well, to justify the cost of everyone involved. As an aside, these PRs also tend to take longer to review and fact-check.
> 
> Hence, @stevewhims, thanks for bringing me up to speed with the concerns behind reviewing and merging certain PRs. I would examine the criteria, should I decide to make another one in the future. Thanks again!

Point taken, so I've accepted this PR. Thanks again! :)

PR #2000, opened 2025-03-19
Remove non-existent pseudo handle

mlindgren:
Removed reference to `BCRYPT_AES_ALG_HANDLE` which doesn't exist


alvinashcraft (approved):
Thanks for addressing this.

PR #2006, opened 2025-04-08
Update using-dynamic-data-exchange.md

lubomirjurukov:
Fix example calls to PackDDElParam and UnPackDDElParam to use UINT_PTR casts instead of UINT so the examples work under 64bit correctly. The corrected examples ware tested with a 64 bit client and server


alvinashcraft (approved):
This is a good update. Thank you!

PR #2008, opened 2025-04-09
Fix COM0 and LPT0 incorrectly listed as reserved

dagood:
* Revert https://github.com/MicrosoftDocs/win32/pull/1415
* Noticed through https://github.com/golang/go/issues/67245
* Confirmed internally that there is no indication that COM0 and LPT0 are intended to be reserved.

/cc @qmuntal


alvinashcraft (approved):
Thanks for the PR, @dagood. Looks good. Do you want to wait for anyone else from the team to review this, or should I go ahead and merge the changes?

dagood:
No problem! Thanks, I think this is good to go ahead and merge. After all, it's only a revert of a disputed change. If evidence for the change is found, it can be reapplied.

It might be reasonable to ask the original author why they made the change, to see if they had some evidence in mind that we simply didn't ask for at the time. In contact with internal owners/experts of this area, we haven't been able to find any evidence. On the other hand, given the PR was merged several years ago, I'm not totally sure it's worth bothering the original author over it.

(The email thread subject is "Are COM0 and LPT0 reserved names?" for any future archaeologists. 🙂)

PR #2010, opened 2025-04-17
Refine context switch description

EricccTaiwan:
Clarify that the scheduler uses separate ready queues per priority level and that threads are preempted or yield the CPU, rather than "finishing execution." Also replaces "load" with "restore" for clarity.

Improve accuracy and readability of the documentation.

### DIFF
--- a/desktop-src/ProcThread/context-switches.md
+++ b/desktop-src/ProcThread/context-switches.md
@@ -1,5 +1,5 @@
 ---
-description: The scheduler maintains a queue of executable threads for each priority level.
+description: The scheduler maintains separate queues of executable threads for each priority level.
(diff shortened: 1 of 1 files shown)


Karl-Bridge-Microsoft (approved):

EricccTaiwan:
Hi @Karl-Bridge-Microsoft, thank you for merging the patch. 
I was wondering—since the update has been merged, will the corresponding documentation page also be updated?

[Context Switches](https://learn.microsoft.com/en-us/windows/win32/procthread/context-switches)

Thanks!

Karl-Bridge-Microsoft:
The updates should be live by now. I'll look into it.

EricccTaiwan:
OK, thanks for your help!

Karl-Bridge-Microsoft:
Looks like the updates are finally live.

PR #2011, opened 2025-04-18
Add the keys for the alt+esc shortcut

binki:
The table of Windows shortcuts and Terminal Services Client shortcuts documents the alt+esc shortcut but did not include the shortcut itself.


GrantMeStrength:
Thanks for the fix! Recreated in the internal repo for merge.

PR #2013, opened 2025-04-25
Replace '❓' with '❔' in DirectX format support

Notheisz57:
Regarding f0d952cd63ba7b7f00c828bdb975e26c6038ea96 authored by @Karl-Bridge-Microsoft.

These pages are invaluable when performing DirectX format/functionality cross-comparisons, to evaluate _best-fit_ formats, etc.

Requesting that we use '❔' (`U+2754`) rather than '❓' (`U+2753`), so that _optional support_ is more clearly discernible from '❌'.


Notheisz57:

Karl-Bridge-Microsoft (approved):

PR #2014, opened 2025-04-25
(exFAT) Fixed likely copy-paste error.

PhucXDoan:
The "Stream Extension directory entry" was referenced despite the fact that the context was under the File Name directory entry; probably a copy-paste error.


PhucXDoan:
Furthermore: the usage of the NoFatChain bit in the GeneralSecondaryFlags for File Name directory entries is unclear. From what I can tell, this bit should also always be zero. If this is the case, the spec should be explicit about that; or better, make GeneralSecondaryFlags reserved for File Name entries.

alvinashcraft (approved):
Yes, it looks like you're correct about this being a copy/paste error. It should reference File Name in this section. Thank you.

PR #2025, opened 2025-05-04
correct SystemResponsiveness rounding and clamp rules

eskezje:
Replaces the outdated sentence that said  
“Note that values that are not evenly divisible by 10 are rounded up to the nearest multiple of 10. A value of 0 is also treated as 10”.

Clarifies the real driver behavior present in every build from
Vista RTM to Windows 11:

10–100 -> rounded down to the nearest lower multiple of 10  
0–9 and >100 -> clamped to 20 %  
100 (or a missing value, default 100) disables MMCSS  
(driver returns STATUS_SERVER_DISABLED, 0xC0000080)

Source: disassembly of mmcss.sys/mmcss.dll versions 6.0.6000.16386 through
10.0.26100.1 shows:

```c
if (value - 10 > 90)      // 0–9 and >100
    effective = 20;       // clamp
else
    effective = 10 * (value / 10);  // integer division floors to lower 10
```
This logic discards the remainder (integer division) and therefore always floors to a lower multiple of 10.

### DIFF
--- a/desktop-src/ProcThread/multimedia-class-scheduler-service.md
+++ b/desktop-src/ProcThread/multimedia-class-scheduler-service.md
@@ -22,7 +22,7 @@ The MMCSS settings are stored in the following registry key:
 
 **HKEY\_LOCAL\_MACHINE\\SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion\\Multimedia\\SystemProfile**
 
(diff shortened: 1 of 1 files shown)


Karl-Bridge-Microsoft (approved):

Lawliet155:
Looks good

PR #2030, opened 2025-05-09
Add missing oplock type names

cgallred:
Updates the description of the link to the "Types of Opportunistic Locks" page.


alvinashcraft (approved):
Thanks for adding these here. Looks good.

PR #2043, opened 2025-05-16
Fix a few typos

Blake-Madden:


desktop-src/WinAuto/variantnotint--checkstate-.md
```diff
@@ -10,7 +10,7 @@ ms.date: 05/31/2018
 
 ## Text
 
-The variant returned from {0} should be an {1} but is a {2}.
+The variant returned from {0} should be an {1}, but is a {2}.
```
alvinashcraft: It's a bit of a nit, but I wouldn't put a comma here. The second half is not a complete sentence. if it were "but it is a {2}", then I would add a comma. Also, the "an" before "{1}" should be "a" instead.

```suggestion
The variant returned from {0} should be a {1} but is a {2}.
```

alvinashcraft (approved):

PR #2057, opened 2025-05-28
Update ioctl-lmr-disable-local-buffering.md

rodwiddowson:
Start pulling a very long string.

Despite the name, this is an FSCTL but of type FILE_DEVICE_NETWORK_FILE_SYSTEM.  This means that if you call DeviceIoControl on it you will get a failure.  You have to send it via NtfsControlFile


rodwiddowson:
Quiet in here.

As it stands the documentation is plain wrong

Karl-Bridge-Microsoft:
Thanks for the suggestion. I'm trying to confirm the details with the product team and will update the topic based on their guidance.

GrantMeStrength:
Thanks for the correction and for your patience! Recreated in the internal repo with cleaned-up wording.

PR #2072, opened 2025-07-08
Update installation-and-configuration-for-windows-remote-management.md

Ch-cyber85:
[CLOSED WITHOUT BEING MERGED; PLEASE SEE MY COMMENT BELOW]

Hello Team,

I am one of the Security Assurance Engineer who handles vulnerability submissions coming to MSRC Windows. There is an MSRC case : 99268 on which we have report where the researcher found a vulnerability with "Set-Item WSMan:\localhost\Service\AllowUnencrypted -Value $true">

As per the engineering teams the feedback here is there is No vulnerability here as the user is allowing the unencrypted traffic.  And also, the doc here should also callout/add a Note to highlight setting the value as True isn't considered Secure.


stevewhims:
Hi, please create a PR against the private repo here. Or contact your writer and/or writing team about this PR.

Thanks!
-Steve

PR #2083, opened 2025-07-23
Update taskschedulerschema-versiontype-simpletype.md

Wolf-SO:
Fix the description of the regular expression: `d+` does not 
stand for *double*, but for a non-empty sequence of decimal digits,
which I simplify here as *number*. Also add an example
that demonstrates the use of multi-digit numbers.


stevewhims (approved):

Wolf-SO:
@stevewhims thanks for merging! Do you happen to know when/if the online public documentation will be updated based on these sources?

PR #2085, opened 2025-07-23
Update Video Processor MFT documentation

mattrwoz:
- Add newly supported video formats.
- Include sections on Color space Transforms, Imaging Transforms, Framerate Transforms to give more detail on what parts of the input and output media types the Video Processor will use.
- Add section for Transform Attributes supported
- D3D12 is now supported
- Renderer effects are now supported
- Add details on Video Processor's use by SVR

Most of these properties have been long supported by the Video Processor.  But a user of the component would not know whether a given attribute would be processed or not.  This PR attempts to make it more clear which attributes on the media types will actually be observed by the Video Processor and thus a conversion will be performed.

### DIFF
--- a/desktop-src/medfound/video-processor-mft.md
+++ b/desktop-src/medfound/video-processor-mft.md
@@ -8,7 +8,7 @@ ms.date: 05/31/2018
 
 # Video Processor MFT
 
-The video processor MFT is a Microsoft Media Foundation transform (MFT) that performs colorspace conversion, video resizing, deinterlacing, frame rate conversion, rotation, cropping, spatial left and right view unpacking, and mirroring.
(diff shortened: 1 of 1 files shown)


drewbatgit:
@mattrwoz Is this ready to go live, or are you still working on it / reviewing it?

mattrwoz:
@drewbatgit this is ready to merge.

drewbatgit (approved):

whiskhub:
Many of the newly referenced interfaces/constants are not yet documented - could these also be updated? For example IMFVideoProcessorControl3, IMFVideoRendererEffectControl, MF_XVP_SAMPLE_LOCK_TIMEOUT, MF_SA_D3D12_AWARE. Thanks!

PR #2086, opened 2025-07-24
Create understanding-pe-signatures.md

ericlaw1979:
To help developers understand that a PE's signature does not cover all bytes of the file, only the code sections.

### DIFF
--- a/desktop-src/SecBP/toc.yml
+++ b/desktop-src/SecBP/toc.yml
@@ -28,3 +28,5 @@
       href: password-threat-assessment.md
     - name: "Threat Mitigation Techniques"
       href: threat-mitigation-techniques.md
+  - name: "Understanding PE Signatures"
+    href: understanding-pe-signatures.md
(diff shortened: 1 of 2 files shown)


alvinashcraft:
Thanks for this, Eric. I'm going to make some minor tweaks just to the markdown formatting. The content looks great.

This should also be added to the toc.yml file in the SecBP folder so it appears in the table of contents on the left, and do you think it makes sense to add a row to link to this topic from the table at the bottom of the [best-practices-for-the-security-apis.md](https://learn.microsoft.com/windows/win32/secbp/best-practices-for-the-security-apis) article at the root of this section?

ericlaw1979:
@alvinashcraft Thanks for your help! After the topic was live, I was going to add links to it from the WinVerifyTrust/WinVerifyTrustEx documentation as well as the page that has the API sample for signature verification. (This particular information is of deep interest to security vendors and not particularly important for general developers).

alvinashcraft (approved):

alvinashcraft:
Thanks for this contribution, @ericlaw1979. We have a process that runs nightly to merge changes to the live branch for publishing. So, this new content will be live late tonight PDT.

PR #2087, opened 2025-07-25
Create docs for new StorageProvider and activity-related properties

amittere:
Adding documentation for new StorageProvider and Activity related properties


alvinashcraft:
Thanks for adding these, @amittere. You'll also need to add them to the `toc.yml` file in the properties folder so they appear in the table of contents on the left side of the Learn pages. Also, whichever folder they're added to in the TOC, you should also add them to the corresponding Topic/Description table in the `*-bumper.md` file for that folder. Those act as indexes for each folder to link to the individual property pages.

@jwmsft may have addition advice/suggestions since he owns this area of the docs. It's the first time I've worked with someone on adding new content here.

jwmsft:
Property docs source files aren't organized into folders, only the bumpers and TOC entries are grouped. The existing StoragePrivider props are in the [Core bumper](https://github.com/MicrosoftDocs/win32/blob/docs/desktop-src/properties/core-bumper.md), which seems to be a catch-all for things that aren't in another group.

- It might make sense to move all the StorageProvider properties out of Core into a new TOC/bumper group of their own. 
- ContentUri probably makes sense to add to Core.
- The 3 Activity props could go in Core or in a new Activities group.

Let me know how you'd like to handle these, and I can help with the TOC updates if you'd like.

alvinashcraft (approved):

jwmsft (approved):

PR #2088, opened 2025-07-29
Create stub for Event log EVTX format

ruslan-tux:
Hi! Is it possible to extend existing Event Log documentation with EVTX format specification/binary layout, including header and event record structures? There is documentation for the legacy format, but seems like it's not supported in the latest Windows versions.

### DIFF
--- a/desktop-src/EventLog/event-log-evtx-format.md
+++ b/desktop-src/EventLog/event-log-evtx-format.md
@@ -1,0 +1,9 @@
+---
+description:
+ms.assetid: 0e591b8f-15fd-4513-b8e2-a30dcce7b903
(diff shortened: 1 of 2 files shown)


ruslan-tux:
@microsoft-github-policy-service agree company="Tuxera"

Karl-Bridge-Microsoft:
Thanks for your contribution!

Per feature team, .evt is deprecated and considered a legacy file format. As such, .evtx files are not recommended and do not need documentation.

"There are plenty of more appropriate file formats for when a log file should be portable to other platforms. When the log needs to be taken off machine, the Event Log offers an export to other file formats like XML or plain text."

PR #2094, opened 2025-08-13
Update example to reference signing discussion

ericlaw1979:
Update the example page to link to discussion of the limits of executable file signing verification.


alvinashcraft (approved):
Looks good. Thanks for the additional information!

PR #2120, opened 2025-10-01
RAS_SERVERS group was in the incorrect category.

HerbertMauerer:
RAS Servers group is not using a builtin domain sid, it is domain+rid 553


alvinashcraft (approved):
Looks good. Thanks for moving this row.

PR #2126, opened 2025-10-06
Fix winioctl.h.h typos

m417z:


alvinashcraft (approved):
Thanks for finding and fixing these typos. Looks good.

PR #2134, opened 2025-10-11
Fix AdjustTokenPrivileges call in privilege sample

forderud:
According to the [AdjustTokenPrivileges](https://learn.microsoft.com/en-us/windows/win32/api/securitybaseapi/nf-securitybaseapi-adjusttokenprivileges) documentation, then the `BufferLength` parameter specifies the size of the `PreviousState` parameter. Since `PreviousState=NULL`, then it follows that `BufferLength` should be 0.

I've already tested the changes locally.


alvinashcraft (approved):
Good catch. Thanks for updating this!

PR #2136, opened 2025-10-15
Fix formatting of desktop & window station access rights

forderud:
The `\_` character sequence  doesn't seem to be displayed correctly within a `<dl>` elment.


alvinashcraft (approved):
Thanks for the update. I think many of these escapes are left over from the HTML conversion from MSDN and aren't needed in the new system and markdown.

PR #2141, opened 2025-11-02
Remove client support attributes from Dwm* APIs

riverar:
These APIs are actively used by Microsoft and others; these support attributes don't make any sense.


stevewhims:
I think we should leave that info. To me, the attributes say that support (by Microsoft Support) ended with Windows 7. And I think that's always going to be a true statement. It's independent of how well any given API/feature functions, and it's independent of how prevalent is the use of any given API/feature.

riverar:
When docs say "End of client support," developers reasonably interpret that as "deprecated" or "don't use this." Perhaps it would be more clear to label these APIs as deprecated to discourage use (as doing so would potentially jeopardize the developer's "Microsoft Support-supportable" state)?

PR #2153, opened 2026-01-05
[BCrypt] GetProperty and SetProperty Identifiers: Add BCRYPT_PUBLIC_KEY_LENGTH + Additional Refactors

ibuprof3n:
Adding the BCRYPT_PUBLIC_KEY_LENGTH identifier, including some additional information for when it applies to ECC curves.

Correcting the string values for some of the KEM identifiers.

### DIFF
--- a/desktop-src/SecCNG/cng-property-identifiers.md
+++ b/desktop-src/SecCNG/cng-property-identifiers.md
@@ -146,7 +146,7 @@ Contains the [initialization vector](/windows/win32/SecGloss/i-gly) (IV) for a k
 <span id="BCRYPT_KEM_SHARED_SECRET_LENGTH"></span><span id="bcrypt_kem_shared_secret_length"></span>**BCRYPT\_KEM\_SHARED\_SECRET\_LENGTH**
 </dt> <dd> <dl> <dt>
 
-L"BCRYPT_KEM_SHARED_SECRET_LENGTH"
+L"KEMSharedSecretLength"
 </dt> <dt>
 
 The size, in bytes, of the shared secret for a key encapsulation mechanism (KEM) key. This data type is a **DWORD**.
@@ -156,10 +156,20 @@ The size, in bytes, of the shared secret for a key encapsulation mechanism (KEM)
 <span id="BCRYPT_KEM_CIPHERTEXT_LENGTH"></span><span id="bcrypt_kem_ciphertext_length"></span>**BCRYPT\_KEM\_CIPHERTEXT\_LENGTH**
 </dt> <dd> <dl> <dt>
 
-L"BCRYPT_KEM_CIPHERTEXT_LENGTH"
+L"KEMCiphertextLength"
 </dt> <dt>
 
 The size, in bytes, of the ciphertext for a key encapsulation mechanism (KEM) key. This data type is a **DWORD**.
+
+</dt> </dl> </dd> <dt>
+
+<span id="BCRYPT_PUBLIC_KEY_LENGTH"></span><span id="bcrypt_public_key_length"></span>**BCRYPT\_PUBLIC\_KEY\_LENGTH**
+</dt> <dd> <dl> <dt>
+
+L"PublicKeyLength"
+</dt> <dt>
+
+The size, in bits, of the public key. For ECC we return the size, in bits, of the field modulus. This data type is a **DWORD**.
 
 </dt> </dl> </dd> <dt>
 
--- a/desktop-src/SecCNG/cng-property-identifiers.md
+++ b/desktop-src/SecCNG/cng-property-identifiers.md
@@ -169,7 +169,11 @@ The size, in bytes, of the ciphertext for a key encapsulation mechanism (KEM) ke
 L"PublicKeyLength"
 </dt> <dt>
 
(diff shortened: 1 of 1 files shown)


desktop-src/SecCNG/cng-property-identifiers.md
```diff
@@ -156,13 +156,23 @@ The size, in bytes, of the shared secret for a key encapsulation mechanism (KEM)
...
 
 </dt> </dl> </dd> <dt>
 
+<span id="BCRYPT_PUBLIC_KEY_LENGTH"></span><span id="bcrypt_public_key_length"></span>**BCRYPT\_PUBLIC\_KEY\_LENGTH**
+</dt> <dd> <dl> <dt>
+
+L"PublicKeyLength"
+</dt> <dt>
+
+The size, in bits, of the public key. For ECC we return the size, in bits, of the field modulus. This data type is a **DWORD**.
```
samuel-lee-msft: Good to call out that ECC is peculiar, might wordsmith to something like:

The size, in bits, of the public key. This data type is a **DWORD**.

This specific size returned varies according to the algorithm. For programmatic use, most callers should use the `pcbResult` output of a [BCryptExportKey](/windows/win32/api/Bcrypt/nf-bcrypt-bcryptexportkey) call.

For most algorithms the returned length corresponds to the bit-size of the serialized public key without any further encoding. For ECC the returned length corresponds to the bit-size of the field modulus (P-256 returns 256 even though the public key is represented by a pair of 256-bit coordinates).

samuel-lee-msft (approved):

desktop-src/SecCNG/cng-property-identifiers.md
ibuprof3n: Ahh yea this is a lot better ahaha

We generally want callers to use the `pcbResult` output of a BCryptExportKey call when in a program right? Would it be worth including a variation of this line at the top of the file instead? I'm not entirely sure how all of the identifiers are used, but a general callout towards the top of this page might be worthwhile if this is the expectation for a lot of the identifiers.

Also, since there's a callout for ECC, it might be helpful to include a callout for KEM since the encapsulation key length is a slightly different terminology than the public key length terminology other algorithms use. I was thinking of something like:

For KEM, the returned length corresponds to the bit-size of the encapsulation key.

samuel-lee-msft: I tend to think Encapsulation key is just a fancy name for public key (in the BCrypt API we call it `BCRYPT_MLKEM_PUBLIC_BLOB`), I don't think it needs to be called out.

The BCryptExportKey comment could apply to the `BCRYPT_KEY_LENGTH`. Not opposed to putting something like this in the description for that property too.

ibuprof3n:
@microsoft-github-policy-service agree company="Microsoft"

desktop-src/SecCNG/cng-property-identifiers.md
```diff
@@ -156,13 +156,27 @@ The size, in bytes, of the shared secret for a key encapsulation mechanism (KEM)
...
 
+<span id="BCRYPT_PUBLIC_KEY_LENGTH"></span><span id="bcrypt_public_key_length"></span>**BCRYPT\_PUBLIC\_KEY\_LENGTH**
+</dt> <dd> <dl> <dt>
+
+L"PublicKeyLength"
+</dt> <dt>
+
+The size, in bits, of the public key. This data type is a **DWORD**.
+
+This specific size returned varies according to the algorithm. For programmatic use, most callers should use the `pcbResult` output of a [BCryptExportKey](https://learn.microsoft.com/en-us/windows/win32/api/bcrypt/nf-bcrypt-bcryptexportkey) call.
```
alvinashcraft: Absolute URLs can cause issues, especially with region info included.
```suggestion
This specific size returned varies according to the algorithm. For programmatic use, most callers should use the `pcbResult` output of a [BCryptExportKey](/windows/win32/api/bcrypt/nf-bcrypt-bcryptexportkey) call.
```

desktop-src/SecCNG/cng-property-identifiers.md
```diff
@@ -171,6 +185,8 @@ L"KeyLength"
 
 The size, in bits, of the key value of a symmetric key provider. This data type is a **DWORD**.
 
+This specific size returned varies according to the algorithm. For programmatic use, most callers should use the `pcbResult` output of a [BCryptExportKey](https://learn.microsoft.com/en-us/windows/win32/api/bcrypt/nf-bcrypt-bcryptexportkey) call.
```
alvinashcraft:
```suggestion
This specific size returned varies according to the algorithm. For programmatic use, most callers should use the `pcbResult` output of a [BCryptExportKey](/windows/win32/api/bcrypt/nf-bcrypt-bcryptexportkey) call.
```

desktop-src/SecCNG/cng-property-identifiers.md
ibuprof3n: Ah thanks for the heads up!

alvinashcraft:
I'm adding @GrantMeStrength to the PR too. I've been moved over to the Microsoft Foundry docs team, so he can assign another writer if needed. The changes look good to me though, as long as @samuel-lee-msft is happy with them.

GrantMeStrength (approved):

PR #2166, opened 2026-03-10
Fixes example

tim-weis:
Adjusts the `helpstring` texts to match the interface effects.

Previously, both `helpstring` texts read "Returns and sets RGB color.", which is inaccurate for either the `propget` or the `propput` method.


tim-weis:
I just realized that the `helpstring` attributes are on `propget`/`propput` members, which often do not translate to individual functions for consumers. Using the same `helpstring` for both does make sense, so I'm closing this PR.

PR #2177, opened 2026-04-23
Fix parameter in sample calls to BCryptSetProperty

jonproch117:
&hKeyPair was being passed instead of hKeyPair, resulting in a failure with STATUS_INVALID_HANDLE when attempting to run the sample code


GrantMeStrength:
Thanks @jonproch117! This fix has been applied to the private authoring repo and will publish with the next sync. Closing this PR.

PR #2181, opened 2026-05-13
WM_CHAR: the menu mode flag

nikita-leontiev:
According to [Keyboard Input Overview](https://learn.microsoft.com/en-us/windows/win32/inputdev/about-keyboard-input), 28 bit is the menu mode flag (KF_MENUMODE mask).
Verified on Windows 7 SP1 x64: 28 bit is actually used for WM_CHAR message and indicates whether a keystroke occurred in menu or not.


DJm00n:
See #2184

GrantMeStrength:
Thank you for this contribution! The fix has been reviewed and applied to the private authoring repository (win32-pr PR #3488). This public mirror PR can be closed.

PR #2183, opened 2026-05-22
Revise extended-key flag details in keyboard input doc

DJm00n:
Updated the extended-key flag section to include additional keys and clarified the distinction of the Num Lock and right-hand Shift keys.


GrantMeStrength:
Thank you for this contribution! The fix has been reviewed and applied to the private authoring repository (win32-pr PR #3488). This public mirror PR can be closed.

PR #2190, opened 2026-06-04
Fix terminology for SECBUFFER_EXTRA handling in Schannel

anrossi:
Clarified the handling of excess input buffers in Schannel by correcting the terminology from 'output buffer' to 'input buffer' for SECBUFFER_EXTRA. `InitializeSecurityContext` and `AcceptSecurityContext` signal extra input buffer through this mechanism; I don't know if `DecryptMessage` does the same.


GrantMeStrength:
Thank you for this report! The SECBUFFER_EXTRA terminology has been corrected — it is indeed the input buffer that gets marked, not an output buffer. Fix is in review and will publish shortly.

PR #2192, opened 2026-06-09
Fix using-resources.md example code

KillerxDBr:
The example code in the section "Creating a Resource List" does not work for unicode builds, since it sends the string size in characters instead of bytes to the WriteFile function, writing incomplete strings and corrupting the output UTF-16 text file

### DIFF
--- a/desktop-src/menurc/using-resources.md
+++ b/desktop-src/menurc/using-resources.md
@@ -193,11 +193,11 @@ if (FAILED(hResult))
     return;
 }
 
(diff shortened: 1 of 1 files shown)


KillerxDBr:

GrantMeStrength:
Thank you for this report! All five WriteFile calls have been corrected to use `cbString * sizeof(TCHAR)` for the byte count. This is an important correctness fix for Unicode builds. The fix is in review and will publish shortly.

PR #2197, opened 2026-07-06
Modernize code examples in using-cursors.md and using-icons.md

DJm00n:
The original examples were written in an older C++ style with filler comments, unused variables, and raw hex byte arrays for cursor/icon masks that were hard to read and error-prone to maintain.

- Cursor and icon mask examples: replaced unreadable hex byte arrays with a symbol map (readable ASCII art) and a PackCursorMasks() helper
- Alpha blended cursor: the original example was incomplete and would not work correctly on modern systems; replaced with a self-contained implementation using analytic AC_SRC_OVER compositing
- Mousetrap: rewrote as a complete, runnable window procedure
- Keyboard cursor movement: fixed incorrect ordering of key check vs GetCursorPos, added acceleration with an upper bound
- Sharing Icon Resources: LoadLibrary with a bare filename is a security issue; replaced with LoadLibraryEx + LOAD_LIBRARY_AS_DATAFILE
- Cleaned up all examples: removed filler comments, unused variable declarations, and outdated C++ style
- Updated all API reference links to /windows/win32/ and wide (W) variants
- Deduplicated GetIconDimensions/GetCursorDimensions between the two files; fixed ownership rules for DestroyIcon

### DIFF
--- a/desktop-src/menurc/using-cursors.md
+++ b/desktop-src/menurc/using-cursors.md
@@ -25,627 +25,486 @@ ms.date: 05/31/2018
 This section discusses the following topics.
 
 -   [Creating a Cursor](#creating-a-cursor)
+-   [Using Cursor Functions to Create a Mousetrap](#using-cursor-functions-to-create-a-mousetrap)
 -   [Creating an Alpha Blended Cursor](#creating-an-alpha-blended-cursor)
 -   [Getting a Cursor size](#getting-a-cursor-size)
 -   [Displaying a Cursor](#displaying-a-cursor)
 -   [Confining a Cursor](#confining-a-cursor)
--   [Using Cursor Functions to Create a Mousetrap](#using-cursor-functions-to-create-a-mousetrap)
 -   [Using the Keyboard to Move the Cursor](#using-the-keyboard-to-move-the-cursor)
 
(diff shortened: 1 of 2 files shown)


GrantMeStrength (approved):
Substantial improvement from someone with strong domain knowledge of these APIs. Standout changes:

- **Security fix**: replacing `LoadLibrary` with a bare DLL name with `LoadLibraryEx(..., LOAD_LIBRARY_AS_DATAFILE)` — the original was a DLL hijacking vector and this is the correct mitigation
- **Alpha-blended cursor**: original example was incomplete and wouldn't run correctly on modern systems; the replacement uses correct AC_SRC_OVER compositing
- **Hex masks → readable symbol map**: a genuine readability improvement for bit-pattern masks
- **Key check ordering fix**: checking key state before calling GetCursorPos is the correct sequence
- **Cursor/icon ownership deduplication**: removing the duplicate GetIconDimensions/GetCursorDimensions helpers reduces maintenance surface

The scope is large but the individual changes are each justified and technically sound.

GrantMeStrength:
@copilot /review

GrantMeStrength (approved):
Reviewed ✅ — Documentation improvement looks good.

PR #2199, opened 2026-07-08
Fix erroneous "HWMD" in function signature

di-gital:


GrantMeStrength:
Thanks for catching this typo, @di-gital! HWMD → HWND is a good catch. I'm going to close this here and apply the fix through our internal repo for preview/publishing. Thanks for the contribution! 🙏

PR #390, opened 2020-06-05
Update context.md

TSlivede:
More "Additional information" links!

The second link I added should possibly be changed to the official documentation, where `%L` `%w`, etc. are explained - but I couldn't find that information in any official microsoft documentation page...

"Allow edits by maintainers" is enabled, so you can change that link to an official documentation page about these <code>%<i>something</i></code> placeholders.

PR #2206, opened 2026-08-01
Update shell/context.md

TSlivede:
Continuation of https://github.com/MicrosoftDocs/win32/pull/390:    
[@GrantMeStrength](https://github.com/GrantMeStrength) [mentioned](https://github.com/MicrosoftDocs/win32/pull/390#issuecomment-4961281223)
> If the change is still relevant, please open a new pull request against the current branch.

------------

More "Additional information" links!

The second link I added should possibly be changed to the official documentation, where `%L` `%w`, etc. are explained - but I couldn't find that information in any official microsoft documentation page...

"Allow edits by maintainers" is enabled, so you can change that link to an official documentation page about these <code>%<i>something</i></code> placeholders.

### DIFF
--- a/desktop-src/shell/context.md
+++ b/desktop-src/shell/context.md
@@ -24,6 +24,8 @@ Additional information is available here:
 
 -   [How To Define Extended Verbs](how-to-define-extended-verbs.md)
 -   [How To Associate Verbs with DDE Commands](how-to-associate-verbs-with-dde-commands.md)
+-   [Verbs and File Associations](fa-verbs.md)
+-   [Old version of this article - The "Community Additions" contain information about possible `%...` replacement tokens](https://web.archive.org/web/20111002101214/http://msdn.microsoft.com/en-us/library/windows/desktop/cc144101(v=vs.85).aspx#contentFeedbackContainer)
 
 ## Shortcut Menus for File System Objects
 


TSlivede:
Regarding

> The topic in question looks very stale

Yes, the documentation was never great, but AFAIK the content is still just as applicable and correct as it was when it was written.

GrantMeStrength:
Thanks for improving the links here, @TSlivede! The addition of the [Verbs and File Associations](fa-verbs.md) link is a good one and I'd like to keep it.

However, we can't include the `web.archive.org` link to the old MSDN page in the published article. Archived third-party snapshots aren't an appropriate reference for Learn documentation, and the accompanying note about "Community Additions" exposes editing/meta commentary to readers rather than documenting the behavior itself.

If the `%...` replacement-token information is valuable (it is!), the right approach is to document those tokens directly in the article as first-class content with a current, citable source, rather than pointing readers to an archived page. Could you update this PR to keep the `fa-verbs.md` link and drop the archive link? I'm marking it `needs-work` so the needed revision is clear. Thanks again for contributing!

GrantMeStrength:
Thank you for the contribution. Review is continuing in MicrosoftDocs/win32-pr#3572 in the private authoring repository, and Copilot review has been requested there. Closing this public PR to keep one review thread. Original authorship and this public discussion are retained in the private PR provenance. The private port includes the Microsoft Learn verbs/file-associations link. It deliberately omits the archived community-comments link because those token details are not an authoritative specification.

PR #2207, opened 2026-08-02
Try to document what happens when a write gets cancelled

Keno:
This PR is partially an attempt to document and partially a complaint that the behavior is so inconsistent in the first place that documentation is required ;). In particular, the various different drivers appear to have all interpreted the existing cancellation guidelines differently, so if this PR were to spur a discussion within the kernel team about how driver are *supposed to* handle this situation, so we could document that in the appropriate driver docs and then only document known exceptions, that would be great :). The docs even have a "call to action" (e.g. in https://learn.microsoft.com/en-us/previous-versions/windows/hardware/design/dn613954(v=vs.85)#call-to-action-and-resources) to adopt cancellation, so I don't think I'm doing anything here that is not intended to work :).

Write cancellation in particular appears to have significantly divergent behavior across different drivers. Both the result status code and whether a cancelled write reports how many bytes were actually written differs significantly between drivers. This PR is the result of QA findings after we recently added (https://github.com/libuv/libuv/commit/6ea389cf64216e3c650eed9197e59cbb1f61c8d5) support for write cancellation to the libuv I/O library (used by e.g. Node.JS, Julia, some Rust applications, etc.). My original understanding was that the written-bytes accounting was reliable on windows, but as mentioned above, this is not true, so this PR is my best effort to document for *which* I/O objects it works, so that I can link this table from the downstream docs. Since this documents kernel behavior, I think it makes most sense to have a table like this here (and as I said, if some of these are considered bugs, that might be even better). The table was populated through some (AI-assisted) empirical testing on Windows Server 2025 Datacenter build 10.0.26100.7171 - it's possible behavior differs on other Windows versions.

### DIFF
--- a/desktop-src/FileIO/canceling-pending-i-o-operations.md
+++ b/desktop-src/FileIO/canceling-pending-i-o-operations.md
@@ -3,7 +3,7 @@ description: Allowing users to cancel I/O requests that are slow or blocked can 
 ms.assetid: adfe6d05-f30b-40a1-b3b0-58e2593e7b25
 title: Canceling Pending I/O Operations
 ms.topic: concept-article
-ms.date: 05/31/2018
+ms.date: 08/02/2026
 ---
 
 # Canceling Pending I/O Operations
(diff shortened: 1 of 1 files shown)


Keno:
@microsoft-github-policy-service agree [company="JuliaHub, Inc."]

Keno:
@microsoft-github-policy-service agree company="JuliaHub, Inc."

GrantMeStrength:
Thanks @Keno — this is a genuinely thoughtful writeup, and documenting the divergence in how partial write success is reported is valuable. The difficulty is that the table describes kernel/driver behavior that (as you note) was derived empirically on a single Windows Server 2025 build and may be version- or driver-specific, and some rows may reflect incidental behavior rather than a guaranteed contract.

That puts it squarely in territory that needs sign-off from the kernel/driver owners before it can be published as documented behavior, so I'm flagging it for human review so the right SMEs can weigh in on which rows are intended contract versus behavior that shouldn't be relied on. Leaving the PR open in the meantime.

GrantMeStrength:
Thank you for the contribution. Review is continuing in MicrosoftDocs/win32-pr#3570 in the private authoring repository, and Copilot review has been requested there. Closing this public PR to keep one review thread. Original authorship and this public discussion are retained in the private PR provenance. The private PR is a draft awaiting subject-matter/format-owner confirmation; this transfer is not approval of the proposed behavior or specification.

PR #2211, opened 2026-08-14
Edit CS reason

nohuto:
A time slice (quantum) expiring doesn't by itself cause a context switch, in `KiQuantumEnd`, the block including the only `KiSwapContext` call only runs when `NextThread` is set:

```c
NextThread = CurrentPrcb->NextThread;
if ( NextThread )
{
```
```c
result = KiSwapContext(v72, NextThread, v103);
```

Means if no other thread is selected, `NextThread` remains null and `KiQuantumEnd` returns without switching.

### DIFF
--- a/desktop-src/ProcThread/context-switches.md
+++ b/desktop-src/ProcThread/context-switches.md
@@ -25,7 +25,7 @@ Until threads that are suspended or blocked become ready to run, the scheduler d
 
 The most common reasons for a context switch are:
 
--   The time slice has elapsed.
+-   The quantum of a thread expired, and another thread with at least the same priority is ready
 -   A thread with a higher priority has become ready to run.
 -   A running thread needs to wait.
 


GrantMeStrength:
Thanks @nohuto — the added condition (a quantum expiring only forces a switch when another ready thread of at least equal priority is selected) is a reasonable refinement, and your `KiQuantumEnd`/`NextThread` reasoning is clear. Since this describes internal scheduler behavior, it needs a human kernel SME to confirm the precise wording before publishing, so I'm flagging it for human review.

One small style note for whoever picks it up: the sibling bullets in that list are full sentences ending in a period, so the new bullet should match that style. Leaving the PR open.

GrantMeStrength:
Thank you for the contribution. This change has already been incorporated into the private authoring repository in MicrosoftDocs/win32-pr#3549 and is present on its main branch. Closing this public PR to avoid duplicate review; no replacement PR is needed. The original contribution remains credited in the private history.

PR #2214, opened 2026-09-04
Update API set conceptual topics for Named Groups and current loader behavior

NeilSandlin:
## Summary

Updates the four API Set conceptual topics under `desktop-src/apiindex`. The
existing pages are broadly sound in structure and intent, so this is an
evolution of them rather than a rewrite — sections were corrected, modernized,
or added, and existing material was preserved where it was still accurate.

The gaps addressed:

- The pages predate **Named Groups**, and describe only versioned contract
  names.
- The account of how the loader resolves an API set reference was incomplete,
  and did not distinguish a successful module load from the availability of a
  particular API.
- There was no guidance on the supported way to detect at run time whether an
  API set is present.
- The umbrella library page did not explain the trade-off between the
  compatibility-oriented libraries and their API set variants.

## Changes by page

| Page | Summary |
|---|---|
| `windows-apisets.md` | Clarifies that a contract name is not a DLL name, describes the `api-` and `ext-` prefixes, and shows how a contract name is used. |
| `api-set-loader-operation.md` | Restructured around how an import reaches an implementation: what resolution requires, what a successful load does not tell you, optional API sets and delay loading, and reverse forwarding. |
| `detect-api-set-availability.md` | Documents choosing the query name, keeping the optional code path reachable, and behavior on earlier versions of Windows. |
| `windows-umbrella-libraries.md` | Describes what an umbrella library covers, lists the available libraries, and explains choosing between a library and its API set variant. |

## Metadata

`ms.date` is updated on all four pages. The `description` front matter is also
updated on three pages (`windows-apisets.md`, `api-set-loader-operation.md`,
`windows-umbrella-libraries.md`) to match the revised content — for example,
the umbrella library description no longer says "static-link library", since
the page now covers the API set variants as well. No other metadata is changed.

## Companion pull request

MicrosoftDocs/sdk-api#2256 — https://github.com/MicrosoftDocs/sdk-api/pull/2256

That pull request adds usage guidance to the three `apiquery2` reference pages.

The two sets are complementary: these conceptual pages describe *when* to query
for availability, and the reference pages define the exact query name forms and
binding requirements. Reviewing them together is recommended.

### DIFF
--- a/desktop-src/apiindex/api-set-loader-operation.md
+++ b/desktop-src/apiindex/api-set-loader-operation.md
@@ -1,8 +1,8 @@
 ---
-description: Describes how Windows uses API sets in loader operations. 
+description: Describes how the Windows library loader resolves an API set reference to the binary that hosts the implementation.
 title: API set loader operation
 ms.topic: reference
-ms.date: 05/31/2023
+ms.date: 09/03/2026
 ---
 
 # API set loader operation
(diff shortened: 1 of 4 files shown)


GrantMeStrength:
Moved to the -pr repo, as this is the recommended location for changed made by internal folks.

PR #2215, opened 2026-09-04
Remove WindowProc example code from wm-endsession.md

zhaotianff:
Replace WindowProc example code with WM_ENDSESSION message definition in wm‑endsession.md。

Refer to the page: https://learn.microsoft.com/en-us/windows/win32/winmsg/wm-ncactivate.

> There are no other Microsoft Learn pages covering the WM_ENDSESSION message.

### DIFF
--- a/desktop-src/Shutdown/wm-endsession.md
+++ b/desktop-src/Shutdown/wm-endsession.md
@@ -14,12 +14,7 @@ A window receives this message through its [**WindowProc**](/windows/win32/api/w
 
 
 ```C++
-LRESULT CALLBACK WindowProc( 
-  HWND hwnd,      // handle to window 
-  UINT uMsg,      // message identifier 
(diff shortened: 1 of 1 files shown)


GrantMeStrength:
Thanks @zhaotianff. I do not think we can take this one as-is. The `WindowProc` signature block is the consistent pattern used across the entire WM_* message reference — it shows readers how the message is delivered to a window — so replacing it on a single page with `#define WM_ENDSESSION 0x0016` moves this page out of step with its siblings (`wm-activate`, `wm-close`, and the rest) without adding equivalent information; the raw numeric message value is not something these reference pages normally surface. If you feel the boilerplate `WindowProc` block should change, that is worth raising as a broader discussion for the whole message reference rather than page-by-page. Marking `needs-work` for now and leaving the PR open.

PR #2216, opened 2026-09-04
Remove WindowProc example code from wm-queryendsession.md

zhaotianff:
Replace WindowProc example code with WM_QUERYENDSESSION message definition in wm-queryendsession.md。

Refer to the page: https://learn.microsoft.com/en-us/windows/win32/winmsg/wm-ncactivate.

### DIFF
--- a/desktop-src/Shutdown/wm-queryendsession.md
+++ b/desktop-src/Shutdown/wm-queryendsession.md
@@ -16,12 +16,7 @@ A window receives this message through its [**WindowProc**](/windows/win32/api/w
 
 
 ```C++
-LRESULT CALLBACK WindowProc( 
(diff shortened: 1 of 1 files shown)


GrantMeStrength:
Thanks @zhaotianff. Same feedback as on the companion `wm-endsession.md` change: the `WindowProc` signature block is the established pattern across the entire WM_* message reference, so swapping it on this page for `#define WM_QUERYENDSESSION 0x0011` makes this page inconsistent with its siblings without conveying equivalent information, and the raw numeric message value is not what these reference pages normally document. If you think the boilerplate `WindowProc` block should be reworked, please raise it as a single broader proposal for the whole message reference rather than one page at a time. Marking `needs-work` for now and leaving the PR open.

PR #2224, opened 2026-09-18
Update windows-data-types.md

ryo-kagawa:
The current definitions for LONGLONG and ULONGLONG are written as follows:
```cpp
#if !defined(_M_IX86)
 typedef __int64 LONGLONG; 
#else
 typedef double LONGLONG;
#endif
```

```cpp
#if !defined(_M_IX86)
 typedef unsigned __int64 ULONGLONG;
#else
 typedef double ULONGLONG;
#endif
```

However, the actual definition in WinNT.h is as follows:
```cpp
//
// __int64 is only supported by 2.0 and later midl.
// __midl is set by the 2.0 midl and not by 1.0 midl.
//

#define _ULONGLONG_
#if (!defined (_MAC) && (!defined(MIDL_PASS) || defined(__midl)) && (!defined(_M_IX86) || (defined(_INTEGRAL_MAX_BITS) && _INTEGRAL_MAX_BITS >= 64)))
typedef __int64 LONGLONG;
typedef unsigned __int64 ULONGLONG;

#define MAXLONGLONG                         (0x7fffffffffffffff)


#else

#if defined(_MAC) && defined(_MAC_INT_64)
typedef __int64 LONGLONG;
typedef unsigned __int64 ULONGLONG;

#define MAXLONGLONG                      (0x7fffffffffffffff)


#else
typedef double LONGLONG;
typedef double ULONGLONG;
#endif //_MAC and int64

#endif
```

Therefore, LONGLONG evaluates to __int64 on both x86 and x64 builds.

To verify this, I conducted a simple test with the following implementation:
```cpp
#include <iostream>
#include <typeinfo>
#include <Windows.h>

#if !defined(_M_IX86)
  typedef __int64 MY_LONGLONG;
#else
  typedef double MY_LONGLONG;
#endif

int main()
{
    std::cout << "   LONGLONG: " << typeid(LONGLONG).name()    << std::endl;
    std::cout << "MY_LONGLONG: " << typeid(MY_LONGLONG).name() << std::endl;
}
```

Test Results
x86:
```
   LONGLONG: __int64
MY_LONGLONG: double
```
x64:
```
   LONGLONG: __int64
MY_LONGLONG: __int64
```

I haven't been able to test it myself, but I believe `double` is only for MIDL 1.0.


ryo-kagawa:

GrantMeStrength:
Thanks @ryo-kagawa — collapsing the raw HTML code table into a clean inline `<code>` line is a nice readability improvement.

One concern before this can go in: the original conditional form documents the actual **WinNT.h** definition, where `LONGLONG` / `ULONGLONG` fall back to `typedef double` on older x86 toolsets that lack native 64-bit integer support. Replacing it with an unconditional `typedef __int64 LONGLONG;` reads more cleanly but drops that platform-dependent detail, so the documented typedef would no longer match what the header actually declares.

Could you keep the conditional definition — just rendered as a clean fenced `cpp` code block rather than the old HTML table — so the readability win is preserved without losing the accurate header signature? Marking `needs-work` for now and leaving the PR open.

ryo-kagawa:
@GrantMeStrength 
Thanks for the review.

> One concern before this can go in: the original conditional form documents the actual WinNT.h definition, where LONGLONG / ULONGLONG fall back to typedef double on older x86 toolsets that lack native 64-bit integer support. Replacing it with an unconditional typedef __int64 LONGLONG; reads more cleanly but drops that platform-dependent detail, so the documented typedef would no longer match what the header actually declares.

No, I don't think the original conditional definition reflected the actual definition in WinNT.h.

```cpp
#include <iostream>
#include <typeinfo>
#include <Windows.h>

#if !defined(_M_IX86)
  typedef __int64 MY_LONGLONG;
#else
  typedef double MY_LONGLONG;
#endif

int main()
{
    std::cout << "   LONGLONG: " << typeid(LONGLONG).name()    << std::endl;
    std::cout << "MY_LONGLONG: " << typeid(MY_LONGLONG).name() << std::endl;
}
```

Test Results
x86:
```
   LONGLONG: __int64
MY_LONGLONG: double
```

x64:
```
   LONGLONG: __int64
MY_LONGLONG: __int64
```

As shown in these results, `LONGLONG` evaluates to `__int64` rather than `double` even on x86 builds.

To faithfully reproduce the definitions in WinNT.h, the content is as follows:
```cpp
#if (!defined (_MAC) && (!defined(MIDL_PASS) || defined(__midl)) && (!defined(_M_IX86) || (defined(_INTEGRAL_MAX_BITS) && _INTEGRAL_MAX_BITS >= 64)))
typedef __int64 LONGLONG;
#else
#if defined(_MAC) && defined(_MAC_INT_64)
typedef __int64 LONGLONG;
#else
typedef double LONGLONG;
#endif
#endif
```

